PR #58, opened 2014-06-21
Allow transition animation overriding in NavigationService.

7Pass:
There are cases when there's a need to override the transition defined in the page. This pull requests adds the option to do so using the UriBuilder.WithTransition.


nigel-sampson:
At the moment I'd like to keep the platform things out of the base navigation services. Apologies for keeping this around for far too long.

PR #94, opened 2014-09-30
PropertyChangedBase.IsNotifying made virtual

mcetkovsky:
The change enables keeping the IsNotifying property value in sync between several objects.

For example, a new value of the IsNotifying property on the parent object can now be propagated into instances contained in properties of this parent object.

Imagine you have a House class which derives from PropertyChangedBase. House has a property of type Door which also derives from PropertyChangedBase. The goal is to enable a scenario that when someone sets House.IsNotifying to false than Door.IsNotifying is also set to false.


aditya-1994:
Good

nigel-sampson:
Resolved in #271

PR #118, opened 2015-01-22
Update .gitignore

demetrisi:


nigel-sampson:
This looks to be a node.js style ignore, not relevant for a .NET project.

PR #140, opened 2015-03-16
Create NuGet packages for Xamarin Android and iOS

brainoffline:
Adds two entries to the NuGet package definition to create packages for iOS and Android


nigel-sampson:
I think you also need to modify https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/nuget/Caliburn.Micro.Core.nuspec to include the new platforms.

Issue #63, opened 2014-06-30
Key input binding support


dellamonica:
It seems to me that binding a command to a specific key (e.g. Enter, Delete) is a fairly common scenario and this project would benefit from offering native support for such bindings. Moreover, there  is already a sample project in
samples / Caliburn.Micro.KeyBinding / 
that could be absorbed by the framework.

JobaDiniz:
+1

stukdev:
Try this gesture cal:Message.Attach="[Gesture Ctrl+Up] = [Action Up()];[Gesture Ctrl+Down]] = [Action Down()]" the action is not called. Maybe is not recognize?

wstaelens:
+1

JobaDiniz:
The good thing about Command in MenuItem is that if the command has a Gesture, then WPF will render the key combination along with the menu text

nigel-sampson:
Given some of the input and the changes expected in `4.0.0` it seems this would only make sense in WPF and would be of dubious value. 

I'm thinking we'll leave the sample "as is" and not add anything to the framework.

ahmediddma:
> 
> 
> Given some of the input and the changes expected in `4.0.0` it seems this would only make sense in WPF and would be of dubious value.
> 
> I'm thinking we'll leave the sample "as is" and not add anything to the framework.

the code cal:Message.Attach="[Key Enter] = [EnterPressed]" does't seem to work for me in 4.0.62-alpha

PR #148, opened 2015-04-06
Added simple gesture trigger example

Fireforge:
I was looking at the Caliburn.Micro samples for keybinding and noticed that while it had code to parse a gesture with keys and modifiers, there was no example of how to use it. Since I needed this functionality, I added an extra Gesture trigger to use the converter to construct a KeyTrigger that has modifiers.

One drawback is that since the KeyTrigger class only supports one Key, it does not make full use of the MultiKeyGesture functionality.

If there is a better way to make use of the MultiKeyGesture/Binding code provided in this sample, please let me know so I can figure out how to use it that way. I'm no expert on the System.Windows.Input or System.Windows.Interactivity APIs at work here.


nigel-sampson:
Thanks, sadly I don't know enough either. There is a standing task to add better key / gesture support with #63. Key Accelerators are being added to Win 10 so real functionality requirement here would just be WPF.

Issue #141, opened 2015-03-23
Windows 10 Universal App Platform


nigel-sampson:
Currently a place-holder issue for commits / branches and discussion about what adding support for the new platforms will look like.

Things to consider
- [x] `Split View` - Will need to modify `BindingScope` to that we examine `SplitView.Pane` for content. Looks like `SplitView` doesn't inherit from `ContentControl` so even more work here.
- [x] API unification - Consider the impacts of unifying the Windows 8.1 and Windows Phone 8.1 APIs.
- [x] Compiled data binding - What impact will this have? Need to find out if we can detected these compiled bindings in some way. `x:Bind` won't be able to be detected, will need to publish guidance of using `x:Bind` in partnership with conventions.
- [x] Device Family folders - Does this effect view location?
- [x] Behaviours SDK is now part of platform
- [ ] ~~Can see some Validation properties on Control, can we use them?~~
- [x] Does the `ISettingsService` make sense now? It's marked as obsolete so won't have code migrated.
- [ ] ~~What do we need to do to support `KeyAccelerator`?~~
- [x] Need to take into account new `SystemNavigationManager`. ([source](http://www.sharpgis.net/post/2015/05/21/Displaying-a-backbutton-in-your-app-window))
- [x] There's a new xaml page template that doesn't have a code behind, it looks like it's not doing a very good job of generating / maintaining the class based on x:View.
- [x] Back button support.
- [x] Remove project.lock.json
- [x] Sample app

unt1tled:
Cool. Was just going to ask about this.

nigel-sampson:
Having some issues with my Windows 10 VM's at the moment but from what I can tell so far the Windows 8.1 version of CM should continue to work with no problems. 

The only issues that may arise are when the CM API surface area between Windows 8.1 and Windows Phone 8.1 differ then these features may not be available right now. I think this is only limited to `NavigationService.BackPressed`.

Will add to this when I know more.

unt1tled:
I was able to add the reference to the UAP project no problem, but I the bootstrapper from the platforms namespace didn't carry over. I started creating a new platform project based on UAP, and indeed, most of the 8.1 stuff are one to one, with the exception a few gotchas.

nigel-sampson:
[...]

PR #149, opened 2015-04-08
Windows 10

ozzeD:


nigel-sampson:
Is there a reason you've created this pull request? Given the state of Windows 10 and especially since we're pre Build conference this stuff is highly likely to change and I'd prefer to keep it out of master.

If you want to discuss the work being done to support Windows 10 there's already an issue raised with #141.

Thanks

Issue #154, opened 2015-04-15
Action conventions for async methods


BrunoJuchli:
The current conventions are like `x:Name="Delete"` ==> Binds to methods `Delete()` and guard property `CanDelete`.

The [guideline](https://msdn.microsoft.com/en-us/library/hh191443.aspx#BKMK_NamingConvention) for methods with `async` modifiers is to add an `Async` suffix to them. For example:

    public async Task DeleteAsync()

Now there's a question:
should the convention pick up `x:Name="Delete"` => `DeleteAsync()` or should we rename to `x:Name="DeleteAsync"`?

Also, the convention should accept guard methods which don't have an `Async` suffix. For example:

    public async Task DeleteAsync();
    public bool CanDelete { get; }

per guideline the property should not have an Async suffix because it's not `async`.

nigel-sampson:
It's certainly something to discuss. I'm ok with what you're proposing here.

Not sure when I'll get to it however :-)

BrunoJuchli:
thanks. If you can give me a pointer where the relevant code is located (roughly) i'll might be able to take a stab at it in the next few days or weeks :)

nigel-sampson:
The code that does the matching is in [ViewModelBinder](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Caliburn.Micro.Platform/ViewModelBinder.cs#L121). 

What I would add is code to check if the return type of the method is assignable to `Task` and the method name ends in "Async". If so then remove the "Async" from the method name. Only if it can't find a correct match on the exact name.

altso:
Is it ok in general to bind event handlers to methods that return `Task`? I see that any exceptions are swallowed if method returns `Task`, but got propagated to `Application.UnhandledException` if method is `void`.

    public Task DeleteAsync() {
        throw new Exception(); // you will never catch this
    }

    public void Delete() {
        throw new Exception(); // triggers CaliburnApplication.OnUnhandledException
    }

I can provide a working example if needed.

BrunoJuchli:
@altso

Methods returning `Task` are treating by Caliburn.Micro like coroutines. When the Task throws an `Exception` you'll be informed by the `Coroutine.Completed` event. The `ResultCompletionEventArgs.Error` property will contain the `Exception`.

nigel-sampson:
[...]

Issue #157, opened 2015-04-22
Clean up Unit Tests


nigel-sampson:
Ensure all the unit tests are able to be run and passing appropriately.

Considering breaking up the unit tests to more appropriate projects / platforms.

Use a .net standard assembly for tests against the .Core projects.

nigel-sampson:
Might be a good idea to just move to MSTest since it has UITest and I don't really have any loyalty to a particular framework.

PR #158, opened 2015-04-25
[wip] Action conventions for async methods

BrunoJuchli:
#154 
In addition to the existing matching, the following now works as well:
**Foo ==> Task FooAsync();**

- [x] implement Action Convention for async methods with `Async` suffix.
 - Wire `View.Foo` to `ViewModel.FooAsync()`
 - Wire guard of `View.FooAsync` to `ViewModel.CanFoo`
- [x] add a sample
- ~~write unit tests~~
- ~~write spec/integration tests~~


BrunoJuchli:
What about unit & spec tests? As far as i can see there are no unit tests for `VIewModelBinder` as of yet.
Also, i couldn't find any integration tests which would instantiate a View and bind it to a view model. I'd like to have done that to check that the old conventions still work but the new one does too. Also in conjunction with guards.
Did i overlook something? Would you like me to write some - or rather not?

nigel-sampson:
The unit tests aren't in a great shape, the latest version of xUnit has disabled the UI thread tests which need to fixed up in #157. Long term this is something that needs to be done a lot better. 

At the moment I'd add something to one of the samples as a bit of a test

BrunoJuchli:
ok thx.
However i've just found that there was a misunderstanding.

You understood that "ViewModel.FooAsync()" should be wired to "View.Foo" (Foo => set by `x:Name=`) --> `ViewModelBinder.cs`.
However what i meant was is that the guard of "View.FooAsync" should be wired to "ViewModel.CanFoo" --> `ActionMessage.cs`.

Now i've implemented both but i'm thinking that the first part is not really necessary. It's only benefit is that it still works if i change between an async implementation (with  `Async` suffix) and a sync implementation (without `Async` suffix). On the downside it's slower and a bit more code to maintain.

The change to `ActionMessage.cs` is more important because it allows me to name a method `FooAsync` but the guard `CanFoo` (without suffix) as is called for by MS coding guidelines.

nigel-sampson:
Ah yeah now I get you, personally I prefer both. Let me think on it and will let you know.

nigel-sampson:
So I think we should include both, just looking over the diffs to keep consistent style and only a few minor things I noticed, some of it isn't your code as I've noticed it's not being consistent in the code base already but may as well kill two birds with one two and not force a merge issue.

- explicit private
- using var where possible
- I'd move the call to `BuildPossibleGuardNames` to inside `TryFindGuardMethod` rather than as a parameter.
- The indentation on `TryFindGuardMethod` looks a bit off.
- I think there's a line ending issue on the xaml file as the diff is a full replace.

Thanks again.

BrunoJuchli:
ok thanks for your feedback. I'll implement the changes as soon as i can get to it :)

BrunoJuchli:
@nigel-sampson 
ok i've implemented the changes and fixes to formatting.
However, i didn't move the call to `BuildPossibleGuardNames` because it's result are used in two places (`PrepareContext`, the method calling it, and `TryFindGuardMethod`).

nigel-sampson:
Thanks, apologies for letting this one slip.

PR #171, opened 2015-05-08
Fixes references to Xamarin.Forms package

altso:
Line 103 caused build errors if Xamarin.Forms 1.3.1.6296 was not present.


nigel-sampson:
Thanks, not sure how that slipped in.

PR #174, opened 2015-05-12
Make sure MainWindow isn't disposed before using it as owner

dvdvorle:
This is a fix for an edge case where code that is not under my control closes the Application.Current.MainWindow (without closing the actual application because it's native c++ in which WPF is hosted). This previously raises an exception when that closed window is used as Owner when creating a new Dialog.

Using PresentationSource.FromVisual as per http://stackoverflow.com/a/26355993/481635


nigel-sampson:
Thanks

PR #186, opened 2015-06-11
Prepare the view frame when displaying a root view for a view model

bendetat:
The Win8 `CaliburnApplication` calls `PrepareViewFirst()` in `DisplayRootView()` but not in `DisplayRootViewFor()`. Meaning that the frame never gets created when displaying a view for a view model as the root view.

This just adds a call to `PrepareViewFirst()` in `DisplayRootViewFor()` where it gets called in `DisplayRootView()`.


bendetat:
Closing this, it turns out that I missed the distinction between view first and view model first. Very confusing.

nigel-sampson:
Thanks though, I'll admit the naming on those methods could be better to highlight the differences between view first and view model first. If you have any thoughts I'd love to hear them.

PR #205, opened 2015-07-24
NameTransformer is faster with compiled Regex

Bobdina:
This is a heavily used codepath. I ran tests, compiled regex is faster (excluding first hit).
The first hit is about 8 times slower, but still a very low number (7 milliseconds on my machine).

### DIFF
--- a/src/Caliburn.Micro.Platform/NameTransformer.cs
+++ b/src/Caliburn.Micro.Platform/NameTransformer.cs
@@ -9,7 +9,7 @@
     /// </summary>
     public class NameTransformer : BindableCollection<NameTransformer.Rule> {
         bool useEagerRuleSelection = true;
-        
+
         /// <summary>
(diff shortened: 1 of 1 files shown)


nigel-sampson:
Thanks, what's the performance difference on the subsequent calls?

Bobdina:
For a 1000 calls, it's about 22ms to 11ms. I made a gist of the testrun I played with, so see for yourself: https://gist.github.com/Bobdina/48c7584627e77a01cf78

nigel-sampson:
Thanks

nigel-sampson:
Just a heads up on this one for next time that this broke the build on all platforms except WPF

Bobdina:
Sorry, how did this break the build exactly?

nigel-sampson:
`RegexOptions.Compiled` isn't a valid enum option in Silverlight and WinRT, but now is in UWP.

PR #225, opened 2015-09-03
Pivot control & Windows UWP

mvermef:
Modified the directive for WP81 to include the WINDOWS_UWP directive, which now has access to the Pivot control in Windows UAP apps.


nigel-sampson:
Thanks

PR #227, opened 2015-09-07
Fixed issue which triggered AmbiguousMatchException

thewindev:
In the ViewLocator for UWP apps, this line of code triggered an AmbiguousMatchException
    var method = element.GetType().GetTypeInfo()
                    .GetDeclaredMethod("InitializeComponent");

In UWP apps, there are two methods called "InitializeComponent" for device specific family views, but one takes a Uri as a parameter, so when you call GetDeclaredMethod it will throw that exception.

I fixed this by calling GetDeclaredMethods() and choosing the first one, which doesn't have the Uri parameter.

### DIFF
--- a/src/Caliburn.Micro.Platform/ViewLocator.cs
+++ b/src/Caliburn.Micro.Platform/ViewLocator.cs
@@ -438,7 +438,7 @@ namespace Caliburn.Micro
         };
 
         /// <summary>
-        ///   When a view does not contain a code-behind file, we need to automatically call InitializeCompoent.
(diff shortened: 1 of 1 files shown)


nigel-sampson:
Thanks for this one, just want to spend some time doing research on when and where this exception shows up as I haven't run into while doing the same thing.

thewindev:
I understand, what I did is just a quick fix. If you need any help in reproducing the error or you need a sample project please let me know.

nigel-sampson:
Thanks for bringing this up, I've committed a fix, pretty much similar to yours but double checks the parameter count to ensure we get the right one.

PR #233, opened 2015-10-03
'transparent' value validation

0x8DEADF00D:
You need to enter the value `transparent` as the background color in the app manifest to make your logos background transparent. If you do so, the settings service will throw a parse exception once you enter a settings flyout. This fix checks if you entered that value and returns a `Transparent` color object.

### DIFF
--- a/src/Caliburn.Micro.Extensions.WIN8/AppManifestHelper.cs
+++ b/src/Caliburn.Micro.Extensions.WIN8/AppManifestHelper.cs
@@ -75,6 +75,11 @@ namespace Caliburn.Micro
 
         private static Windows.UI.Color ToColor(string hexValue)
         {
+            // if 'transparent' is entered in the app manifest, return Windows.UI.Colors.Transparent
+            // in order to prevent parsing failures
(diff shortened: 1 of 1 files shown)


src/Caliburn.Micro.Extensions.WIN8/AppManifestHelper.cs
```diff
@@ -75,6 +75,11 @@ public Windows.UI.Color BackgroundColor
 
         private static Windows.UI.Color ToColor(string hexValue)
         {
+            // if 'transparent' is entered in the app manifest, return Windows.UI.Colors.Transparent
+            // in order to prevent parsing failures
+            if (hexValue == "transparent")
```
altso: You might want to make comparison case insensitive to handle `Transparent` value.

nigel-sampson:
Thanks for the fix, if you can make the comparison case insensitive I'll merge it in.

0x8DEADF00D:
done, thank you for this great framework.

nigel-sampson:
Thanks

PR #243, opened 2015-10-30
Add Debug and Error logs

Maronus:


nigel-sampson:
Thanks for the pull request, can you clarify a bit more what your reasoning in adding these is?

Maronus:
Sorry for my bad English.
Most loggers support more log levels (for example - NLog (https://github.com/NLog/NLog/wiki/Tutorial#log-levels), log4net (https://logging.apache.org/log4net/release/sdk/log4net.Core.Level.html) or MetroLog (https://github.com/onovotny/MetroLog/blob/master/MetroLog.WinRT/ILogger.cs).
I would like to add the remaining.

nigel-sampson:
Thanks for the update, correct me if my misunderstanding is wrong, but you're looking to use Caliburn.Micro's logging infrastructure throughout your app backed by one of the other loggers and want all the levels passed through?

nigel-sampson:
After some thought I don't think we should add extra log levels in.

We're not looking to have a full log framework in Caliburn.Micro. Enough for very small apps to get by and a way for the framework itself to log what's going on.

If you are looking to use other logging frameworks in conjunction I'd recommend implementing an adapter between the two so that CM can log to that framework while your app used that framework directly.

Thanks again.

PR #282, opened 2016-01-12
virtual keyword for event

### LINKED ISSUE #279: Could you change PropertyChangedBase.PropertyChanged to virtual?
Hi,

As most people who already dealt with memory leaks in c# it is known that the most common cause of this is event handlers, which under certain circumstances retain a reference and keep it from being collected by the GC. 

Since i don't want to ask you to change all event handlers to a weak referenced implementation, i thought that you might consider changing the event itself to virtual, so it's possibly to override the implementation to a weak one in a custom derived class.

If changing the implementation to weak referenced handlers would be an option i could share the implementation i've used in https://www.nuget.org/packages/WeakEvent/  (which is PCL compatible)

This implementations performance is 7x that of ordinary events, but offers the convencience of not having to carefully attach/detach handlers in order to get a non leaking software.

Would one of those 2 options be acceptable to do for you?

taori:
This pullrequest partially resolves #279 (changing other event candidates as seperate pull request so you can decide on wether you want those changes or not)

### DIFF
--- a/src/Caliburn.Micro/ConductorBase.cs
+++ b/src/Caliburn.Micro/ConductorBase.cs
@@ -34,7 +34,7 @@
         /// <summary>
         /// Occurs when an activation request is processed.
         /// </summary>
-        public event EventHandler<ActivationProcessedEventArgs> ActivationProcessed = delegate { };
+        public virtual event EventHandler<ActivationProcessedEventArgs> ActivationProcessed = delegate { };
 
         /// <summary>
         /// Gets the children.
@@ -61,17 +61,22 @@
         /// <param name="item">The item on which activation was attempted.</param>
         /// <param name="success">if set to <c>true</c> activation was successful.</param>
         protected virtual void OnActivationProcessed(T item, bool success) {
-            if (item == null) {
+	        if (item == null) {
                 return;
             }
 
-            ActivationProcessed(this, new ActivationProcessedEventArgs {
-                Item = item,
-                Success = success
-            });
+	        var handler = ActivationProcessed;
+	        if (handler != null)
+			{
+				handler(this, new ActivationProcessedEventArgs
+				{
+					Item = item,
+					Success = success
+				});
+			}
         }
 
-        /// <summary>
+	    /// <summary>
         /// Ensures that an item is ready to be activated.
         /// </summary>
         /// <param name="newItem">The item that is about to be activated.</param>
--- a/src/Caliburn.Micro/ConductorBase.cs
+++ b/src/Caliburn.Micro/ConductorBase.cs
@@ -7,24 +7,25 @@
     /// A base class for various implementations of <see cref="IConductor"/>.
     /// </summary>
     /// <typeparam name="T">The type that is being conducted.</typeparam>
-    public abstract class ConductorBase<T> : Screen, IConductor, IParent<T> where T: class {
+    public abstract class ConductorBase<T> : Screen, IConductor, IParent<T> where T : class {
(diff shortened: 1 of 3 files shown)


taori:
if you want me to remove 869e436 i can do so of course

nigel-sampson:
Thanks for the pull request, congrats on your first one.

I'm happy to pull in the changes for the other classes, the only thing I'd change is the extra things you added such as some odd formatting and the checking of the event handler for null.

By declaring the event like `public event EventHandler<ActivationProcessedEventArgs> ActivationProcessed = delegate { };` rather than just `public event EventHandler<ActivationProcessedEventArgs> ActivationProcessed;` we're essentially adding a no-op handler right from the start and means that `ActivatationProcessed` will never be null.

You should just able to do another commit to your fork fixing these up and the pull request will update automatically.

taori:
Thank you :)

Actually i completely overlooked that you're using opening brace at the end of the line. I'll fix the formatting. My VS is just set to MS default code guidelines. But since it's your codebase i'll adapt :)

Regarding the way you're invoking events : Yes, i've read about a variety of methods to do so. While there's only one proper downside to this approach in this case i've only followed resharpers hint. 

This one: 
image

The changes i've made resolved the indicated issue. Thinking about it makes perfect sense too. While the default implementation requires no null check, an overriden event might require checking. At least that's what i can imagine as potential issue.

An alternative would be: ActivationProcessed?.Invoke(this, new ActivationProcessedEventArgs());

However i am still puzzled as to how this compiles for net40. It's as if VS2015 powered by roslyn does compile time source changes. c#6 language features for net40. magic.

I'll change the format, while you decide on this. (But to me it seems getting rid of the hint seems to be essential)

taori:
Actually the odd formatting you were talking about probably also includes me using tabs instead of 4 spaces as 1 tab. I'll have a look if github somehow offers a setting to convert that, without me having to change my IDE

taori:
@nigel-sampson formatting looks almost the same now. using resharper CodeCleanup was what i was missing. i usually use the format of VS. However the shared DotSettings file seems to have one mistake if i am not mistaken?

image

Right now code format force starts a new line instead of collapsing the bracket to a oneliner:
Like here: https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Caliburn.Micro/Screen.cs#L112

. I suppose this formatting is intentional and should therefore be included in the DotSettings file (set to "Place braces together")?

taori:
@nigel-sampson Can i have feedback please on this one: "I suppose this formatting is intentional and should therefore be included in the DotSettings file (set to "Place braces together")?"

nigel-sampson:
Formatting stuff is all good now, I wouldn't worry about it too much beyond what's already done.

The "javascript" style formatting was something I've inherited and far down this list of something I'd like to clean up. 

Regarding the compiler / framework stuff. Yeah a lot of C# features like extension methods are just compiler "tricks" which can work on lower platforms. 

I've yet to include C# 6 stuff in here yet but that'd probably come soon. Bit busy today but will merge this is soon.

taori:
Ok good. Yeah, I don't particularly mind the javascript like style. It's just an odd thing in the C# world. Luckily thanks to resharper fixing it will be an easy fix. hehe.

Yeah true, but in IL i've always noticed how Extension Methods are compiled as static method calls. However i was surprised at first that i was able to use nameof without upgrading my project. 

Anyway: Judging by the resharper warning the invocation change of the event makes sense now and can stay as it is? or are there any more issues with this PR? :)

PR #335, opened 2016-06-04
NavigationPageAdapter now takes ViewModel type into account when pushing view

Nimgoble:
PushAsync now takes the ViewModel type as an argument (defaults to
null).  If the viewmodel type is specified, it will use IoC to get an
instance of it, as opposed to relying on
ViewModelLocator.LocateForView(which ultimately just uses IoC, anyways).
No reason to look up the viewmodel type again, when we already know it.

### DIFF
--- a/src/Caliburn.Micro.Platform/xforms/NavigationPageAdapter.cs
+++ b/src/Caliburn.Micro.Platform/xforms/NavigationPageAdapter.cs
@@ -85,7 +85,7 @@
 
             var view = ViewLocator.LocateForModelType(viewModelType, null, null);
 
-            return PushAsync(view, parameter, animated);
+            return PushAsync(view, parameter, animated, viewModelType);
         }
 
         /// <summary>
@@ -135,14 +135,14 @@
             return NavigateToViewAsync(typeof(T), parameter, animated);
         }
 
-        private Task PushAsync(Element view, object parameter, bool animated)
+        private Task PushAsync(Element view, object parameter, bool animated, Type viewModelType = null)
         {
             var page = view as Page;
 
             if (page == null)
                 throw new NotSupportedException(String.Format("{0} does not inherit from {1}.", view.GetType(), typeof(Page)));
 
-            var viewModel = ViewModelLocator.LocateForView(view);
+            var viewModel = viewModelType == null ? ViewModelLocator.LocateForView(view) : IoC.GetInstance(viewModelType, null);
 
             if (viewModel != null) {
                 TryInjectParameters(viewModel, parameter);


nigel-sampson:
Thanks for the PR

Is there a reason you can't use `navigationService.For<TViewModel>().Navigate();` to avoid specifying the view.

One of the key features of CM is the convention based view / view model location which this sidesteps. Just trying to understand the reasoning why it would be required.

Nimgoble:
Hey Nigel,

Well, I am using that.  But, if you look deeper in the code, that winds up being

`NavigateHelper<TViewModel>().AttachTo(navigationService).Navigate();`

which eventually comes to:

`navigationService.NavigateToViewModelAsync<TViewModel>(parameters, animated);`

If your navigation service is the NavigationPageAdapter, then it's eventually calling `public Task NavigateToViewModelAsync(Type viewModelType, object parameter = null, bool animated = true)`

In that method (NagivationPageAdapter.cs, line 81), it's looking up the view:

`var view = ViewLocator.LocateForModelType(viewModelType, null, null);

return PushAsync(view, parameter, animated, viewModelType);`

In PushAsync, you're calling `ViewModelLocator.LocateForView(view)`, which is looking up the view's corresponding viewmodel type.  This is an unnecessary lookup, since we have already passed the viewmodel type in our original call.  

So, my changes eliminate that look up and, instead, go directly to instantiating the viewmodel type via IoC, which is what would happen in `ViewModelLocator.LocateForViewType`, anyways.

nigel-sampson:
Besides eliminating the convention based lookup is there a scenario you're using where that lookup is incorrect.

I'm looking at bring the navigation services into alignment with the view of an interface that can be used cross platform.

However the Xamarin.Forms one is different than the others in that the view is directly created rather than by the underlying control. So the scenario you have above (specifying both view and view model) only works here bringing them further out out alignment.

Nimgoble:
It's redundant in all scenarios where you navigate by viewmodel.  As I said, there's no reason to look up the viewmodel's class when you're passing it in in the first place(unless there's something that I've missed).  

This doesn't really affect you making the navigation services cross platform, unless you intend to eliminate navigating by viewmodel, right?

Anyways, it's just a small change that I figured would save some time be eliminating a redundant lookup.

nigel-sampson:
So as explained Xamarin.Forms is special in that navigation is done by view instance. All the other platforms follow a pattern similar to the one below.

On all other platforms the control itself doesn't let you navigate by view or view model, just by the view type (or some instances the uri to the view). This leads to the navigation service having to do the following: 

1. Tell the frame to navigate to view `frame.Navigate(typeof(ProductView));`.
2. Frame creates the view and fires Navigated event.
3. Navigation service responds to event, locates view model and binds to it.

View navigation isn't want we want (in my opinion) in our view models so we have things like `.For<T>()`. So we can provide the view model type, which gets transformed the view type to kick off the workflow above.

So while as you said a redundant look up in Xamarin.Forms it's not a scenario that can be provided on other platforms and would prefer to be consistent.

PR #350, opened 2016-07-25
[WIP] Adapt ContentViews to ContentPages during navigation

shaunmarx:
This change allows the `NavigationPageAdapter` to also use a `ContentView` when navigating. I found Xamarin to be a bit too restrictive around it's usage of these for no real reason.

FWIW this allows the same top level view to be either composed as part of a tab/master detail or be navigated to using the navigation service without actually having to change the object type. I wasn't too sure whether this behavior should be baked into the `NavigationPageAdapter` or whether an additional extension point should be added allowing the user to intercept the view and modify it right before it is bound, but happy to make those changes as well if you think it would be worthwhile.

### DIFF
--- a/src/Caliburn.Micro.Platform/xforms/NavigationPageAdapter.cs
+++ b/src/Caliburn.Micro.Platform/xforms/NavigationPageAdapter.cs
@@ -139,8 +139,8 @@
         {
             var page = view as Page;
 
-            if (page == null)
-                throw new NotSupportedException(String.Format("{0} does not inherit from {1}.", view.GetType(), typeof(Page)));
+            if (page == null && !(view is ContentView))
+                throw new NotSupportedException(String.Format("{0} does not inherit from either {1} or {2}.", view.GetType(), typeof(Page), typeof(ContentView)));
 
             var viewModel = ViewModelLocator.LocateForView(view);
 
@@ -148,6 +148,11 @@
                 TryInjectParameters(viewModel, parameter);
 
                 ViewModelBinder.Bind(viewModel, view, null);
+            }
+
+            if (view is ContentView)
+            {
+                page = new ContentPage { Content = (ContentView)view };
             }
 
             page.Appearing += (s, e) => ActivateView(page);
--- a/src/Caliburn.Micro.Platform/xforms/NavigationPageAdapter.cs
+++ b/src/Caliburn.Micro.Platform/xforms/NavigationPageAdapter.cs
@@ -19,6 +19,26 @@
             this.navigationPage = navigationPage;
         }
 
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


shaunmarx:
I think the title needs to be provided to the adapted detail page so that navigation will display the appropriate bits. I think I would honestly prefer extension points to allow us to do this as the current `NavigationPageAdapter` simply doesn't allow for it. 

@nigel-sampson do you believe this is worthwhile adding?

nigel-sampson:
@shaunmarx I think it's useful, sorry this has sat on the backlog.

Just wondering if there's anything else that needs to be added to round this out.

One concern is there would need to be a way to customize the `ContentPage` so you can set things like the `Title`

Maybe extract `page = new ContentPage { Content = (ContentView)view };` out to it's own method.

``` csharp
protected virtual ContentPage CreateContentPage(ContentView view, object viewModel)
{
 ....
}
```

shaunmarx:
@nigel-sampson that is no problem at all. I figured you were busy ;)

I should have a bit of free time to look at this today to implement the changes you have suggested.

shaunmarx:
@nigel-sampson ready for you to have a look at

src/Caliburn.Micro.Platform/xforms/NavigationPageAdapter.cs
```diff
@@ -19,6 +19,26 @@ public class NavigationPageAdapter : INavigationService {
...
+        /// <param name="view">The view to be adapted</param>
+        /// <param name="viewModel">The view model which is bound to the view</param>
+        /// <returns>The adapted ContentPage</returns>
+        protected virtual ContentPage CreateContentPage(ContentView view, object viewModel)
+        {
+            var page = new ContentPage { Content = view };
+
+            var hasDiplayName = viewModel as IHaveDisplayName;
+            if (hasDiplayName != null) {
+                page.Title = hasDiplayName.DisplayName;
```
altso: Not sure if it's applicable to Xamarin Forms, but should `Title` be binded to `DisplayName` to reflect any future changes of the property?

nigel-sampson:
Yeah I think a binding may be a better idea. Should be able to use `Caliburn.Micro.Xamarin.Forms.ConventionManager.SetBinding ` to help.

shaunmarx:
@nigel-sampson, @altso I have added a binding, however I had a look at what happens with this as part of a `NavigationPage` and it would seem that the Navbar provided by Xamarin doesn't seem to automatically update when the Title of `ContentPage` is updated. It only seems to get updated when you navigate away from the current page. I must either be missing something or this is a bug with Xamarin Forms.

By the way, this behaviour seems to be the same even when manually setting the Title for both the ContentPage and parent NavigationPage.

Any ideas?

PR #370, opened 2016-09-14
OnUIThreadAsync() shutdown deadlock under .NET4 fix

smischke:
This pull request fixes a deadlock that occurs under .NET 4.0 when the application is shut down while "XamlPlatformProvider.OnUIThreadAsync().Wait()" is called. The calling thread might stay in "Wait()" forever because the dispatcher operation that would wake it up via "taskSource.SetResult(null)" or "taskSource.SetException(ex)" gets aborted due to the shutdown. If the blocked thread is not a backgound thread, it will prevent the application from shutting down.

"XamlPlatformProvider.OnUIThreadAsync().Wait()" was called in an earlier implementation of "Execute.OnUIThread()" which is used by "PropertyChangedBase.NotifyOfPropertyChange()". The deadlock is not so likely when using CM3 but still possible. 

There might be the same problem in the silverlight version of "XamlPlatformProvider.OnUIThread()" ("waitHandle.Set()" is never called), but I don't use silverlight and have no solution for it.

### DIFF
--- a/src/Caliburn.Micro.Platform/XamlPlatformProvider.cs
+++ b/src/Caliburn.Micro.Platform/XamlPlatformProvider.cs
@@ -20,7 +20,7 @@
 #else
         private Dispatcher dispatcher;
 #if !NET45
-        private static TaskCompletionSource<object> taskForce;
+        private TaskCompletionSource<object> taskSource;
 #endif
 #endif
 
@@ -37,21 +37,21 @@
 #if !NET45
             dispatcher.ShutdownFinished += (sender, args) =>
             {
-                var taskSource = taskForce;
-                if (taskSource != null)
+                var ts = taskSource;
+                if (ts != null)
                 {
-                    taskForce = null;
-                    taskSource.SetResult(null);
+                    taskSource = null;
+                    ts.SetResult(null);
                 }
             };
 #endif
 #endif
     }
 
-    /// <summary>
-    /// Indicates whether or not the framework is in design-time mode.
-    /// </summary>
-    public bool InDesignMode {
+        /// <summary>
+        /// Indicates whether or not the framework is in design-time mode.
+        /// </summary>
+        public bool InDesignMode {
             get { return View.InDesignMode; }
         }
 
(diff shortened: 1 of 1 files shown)


src/Caliburn.Micro.Platform/XamlPlatformProvider.cs
```diff
@@ -19,25 +19,39 @@ public class XamlPlatformProvider : IPlatformProvider {
         private CoreDispatcher dispatcher;
 #else
         private Dispatcher dispatcher;
+#if !NET45
+        private static TaskCompletionSource<object> taskForce;
```
nigel-sampson: I don't think this needs to be static given there's only once instance of the `XamlPlatformProvider`.

src/Caliburn.Micro.Platform/XamlPlatformProvider.cs
```diff
@@ -19,25 +19,39 @@ public class XamlPlatformProvider : IPlatformProvider {
...
 #endif
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="XamlPlatformProvider"/> class.
-        /// </summary>
-        public XamlPlatformProvider() {
+    /// <summary>
+    /// Initializes a new instance of the <see cref="XamlPlatformProvider"/> class.
+    /// </summary>
+    public XamlPlatformProvider() {
```
nigel-sampson: I don't think the indentation changes are needed.

src/Caliburn.Micro.Platform/XamlPlatformProvider.cs
```diff
@@ -19,25 +19,39 @@ public class XamlPlatformProvider : IPlatformProvider {
...
+    }
 
-        /// <summary>
-        /// Indicates whether or not the framework is in design-time mode.
-        /// </summary>
-        public bool InDesignMode {
+    /// <summary>
+    /// Indicates whether or not the framework is in design-time mode.
+    /// </summary>
+    public bool InDesignMode {
```
nigel-sampson: I don't think the indentation changes are needed.

smischke:
Tested under .NET4 using CM 1.5.2. Not sure if it's bullet proof, especially with concurrent calls, but it seems so, because it works in production applications. Not tested with newer CM versions since in our newer applications, where we use CM 3, we also use .NET45 only.

nigel-sampson:
Thanks for the update, realized on another read (and as you mention about concurrent calls) there could be some concerns about multiple calls into `OnUIThreadAsync` from different threads at the same time, in this case stomping over top of each other in terms of the instance `taskSource`.

In your situation with the deadlock what was happening if you "awaited" the method rather than using `.Wait()`?

smischke:
To be honest, I don't know. We use CM 1.5.2 coroutines and not the async await language feature in our old software for legacy reasons. In our new software, we upgraded to CM 3 and C# language level 5 is no problem, but we also use .NET45 where the problem doesn't exist.

I don't know how to solve it in an elegant way. lock()-ing on something like 
`private Dictionary<int, Stack<TaskCompletionSource<object>>> taskSources;`
where "int" is ManagedThreadId and "Stack" is for recursion seems quite clunky.

nigel-sampson:
Thanks, let me know where you get to.

KasperSK:
Closing this as there has not been any work on it for a very long time.

Issue #375, opened 2016-10-02
Enable SimpleContainer to resolve multiple parameters of same type


jongleur1983:
The SimpleContainer is what it's called: quite simple.
Nevertheless I propose to extend it slightly to better support more than one parameter of the same type.
Consider a constructor with the following signature

```public Point(int x, int y)```

Without knowing the internal implementation it's clear that x and y may have to be different values. Currently it's possible to register one instance per type (e.g. int) without a key, or different ones with different keys, where a key is a string value.
These keys are not considered by the constructor resolution, yet.

I propose to change this as follows:
For each constructor argument look for registered instances of matching type. If there's one where the key matches the arguments name, take it, else take another (as it has been done before).

**Comment by nigel-sampson** [comment]
Not sold on this at the moment.

I feel like we're straying away from **Simple** with this change and that if this sort of functionality is desired then a more fully featured container is the better path to go down.

I don't want to bit by bit add more and more functionality to `SimpleContainer` and find ourselves in a years time supporting a fully featured container.

But having said that we have to draw the line somewhere, is it here?

**Comment by jongleur1983** [comment]
Good point.
As I wrote before I never saw SimpleContainer in real use, but I consider constructors with two parameters of the same type a common task.

For Caliburn, where we mostly talk about constructing ViewModels we had (although using another DI container) several cases of "comparison view models" at work, where some screen compares two items of the same type.

Perhaps SimpleContainer is in fact not useful for many scenarios - and probably you're right and it shouldn't be, but in that case I don't get why it's implemented at all yet (no offense intended), as real scenarios require something more sophisticated, and easy scenarios do not require DI at all.

**Comment by nigel-sampson** [comment]
It depends a lot on where you're using the container. Mentally I tend to divide view models into two distinct groups.

1. Page view models, usually inherit from `Screen` and their dependencies are normally application services behind interfaces.
2. Model view models, usually inherit from `PropertyChangedBase` and their dependencies are normally model objects they represent.  These one's don't tend to be created by the container but in the form of code such as:

``` csharp
var products = productsService.GetOnSaleProducts().Select(p => new ProductListViewModel(p));
```

With this approach I don't tend to have multiple values of the same type in the container and so `SimpleContainer` works well.

Would love you to expand on your example as it's not something I'd considered.

jongleur1983:
My example is only a slight variation of yours, splitting your first group into two subgroups:

1. "leaf" screens - that's what I think you had in mind
[...]

PR #376, opened 2016-10-02
#375 support multiple parameters of the same type

jongleur1983:
if there's more than one parameter registered for a given type, with this commit the instance is preferred, where the registered key matches the argument name in the constructor.


nigel-sampson:
Closing after discussion in #375

PR #401, opened 2016-12-29
Data template selector for Xamarin Froms

k-shan:
This PR is regarding adding a DataTemplateSelector that will resolve views based on the viewModel and wire them together (Xamarin.Froms has added DataTemplateSelectors as of verison 2.1.0 ).

Background: Building an app using Xamarin.Forms.TabbedPage resulted in adding child views in XAML and wiring them up manually. Created this DataTemplateSelector to help with that and it works well with the conductors. Thanks

### DIFF
--- a/src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.Xamarin.Forms.csproj
+++ b/src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.Xamarin.Forms.csproj
@@ -49,6 +49,7 @@
     <Compile Include="Parser.cs" />
     <Compile Include="xforms\ActionMessage.cs" />
     <Compile Include="xforms\AttachedCollection.cs" />
+    <Compile Include="xforms\ContextAwareDataTemplateSelector.cs" />
     <Compile Include="xforms\ConventionManager.cs" />
     <Compile Include="xforms\HttpUtility.cs" />
     <Compile Include="xforms\IAttachedObject.cs" />
--- a/src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.Xamarin.Forms.csproj
+++ b/src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.Xamarin.Forms.csproj
@@ -88,15 +88,15 @@
   </ItemGroup>
   <ItemGroup>
     <Reference Include="Xamarin.Forms.Core, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Xamarin.Forms.2.0.1.6505\lib\portable-win+net45+wp80+win81+wpa81+MonoAndroid10+MonoTouch10+Xamarin.iOS10\Xamarin.Forms.Core.dll</HintPath>
+      <HintPath>..\packages\Xamarin.Forms.2.3.3.175\lib\portable-win+net45+wp80+win81+wpa81+MonoAndroid10+Xamarin.iOS10+xamarinmac20\Xamarin.Forms.Core.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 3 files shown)


nigel-sampson:
Thanks for the PR, traditionally we've stayed away from Template Selectors since Caliburn.Micro already has this built in via `View.Model`.

Having a `DataTemplate` the same as the one in [this sample](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/samples/samples/features/Features.Forms/Features.Forms/Views/BindingsView.xaml#L14) has much the same effect as the `ContextAwareDataTemplateSelector`.

We actually expose this on `ConventionManager.DefaultItemTemplate`, normally we'd apply this to ListViews by convention but can't be done in Xamarin.Forms.

k-shan:
Thanks for the feedback.

Using a ContentView in a ViewCell works for ListViews & TableViews. In the case of a [TabbedPage](https://developer.xamarin.com/api/type/Xamarin.Forms.TabbedPage/) the children are of type Page. Xamarin uses the Title (and Icon in iOS) properties on the Page for the tabs - it may be doing other things as well. 

The problem I'm facing is that the object tree structure in Xamarin Forms is limiting. Both Page and View inherit directly from VisualElement meaning we cannot assign a Page to [ContentView.View](https://developer.xamarin.com/api/property/Xamarin.Forms.ContentView.Content/).

So I tried this out:
```
<DataTemplate>
    <ContentPage cm:View.Model="{Binding}" />
</DataTemplate>
```

In the XAML above I believe Caliburn.Micro is attempting to set the `ContentPage.Content` property (content is `View.DefaultContentPropertyName`) however it's type is again View.

If we still want to stay away from Template Selectors do you have any suggestions? Could we change the [OnModelChanged](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Caliburn.Micro.Platform/View.cs#L307) code (or somewhere further down the stack) to accommodate for Xamarin.Forms.Page.

nigel-sampson:
Ah good point, I had forgotten about the difference of `TabbedPage`.

As a work around does the following work.

``` xml
<DataTemplate>
  <ContentPage>
      <ContentView cm:View.Model="{Binding}" />
  </ContentPage>
</DataTemplate>
``

**Comment by k-shan** [comment]
Nope, same issue - getting an ArgumentException. We're trying to set a Page to a View.

**Comment by nigel-sampson** [comment]
So the view being located and being added to the `ContentView` is a `ContentPage`?

**Comment by k-shan** [comment]
Yeah, the view is located fine, its just having issues setting the content property (in this case should be adding it to the TabbedPage)

**Comment by nigel-sampson** [comment]
What you may need to change to use data template from my post is that the views being inserted should inherit from `ContentView` and not `ContentPage`.

**Comment by k-shan** [comment]
I tried that as well, but we then lose out on the Title & Icon properties on the [Page](https://developer.xamarin.com/api/type/Xamarin.Forms.Page/#memberlist) which is used for the tab label/icon.

**Comment by nigel-sampson** [comment]
But no assignment errors? 

You could specify them on the view model representing the tab (the view model already has `DisplayName` for instance) and do something like

``` xml
<DataTemplate>
  <ContentPage Title="[Binding DisplayName}">
      <ContentView cm:View.Model="{Binding}" />
  </ContentPage>
</DataTemplate>
```

nigel-sampson:
Thinking about this long term it may be that for `TabbedPage` a data template selector is the best approach since it has restrictions around `ContentPage`.

Incredibly frustrating.

k-shan:
Just tested this out and works great - I added a Title & Icon property to the ViewModel itself. We could do the same for the other properties on Page as well if needed.
````
<DataTemplate>
	<ContentPage Title="{Binding Title}" Icon="{Binding Icon}" >
		<ContentView cm:View.Model="{Binding}" />
	</ContentPage>
</DataTemplate>
````

nigel-sampson:
Excellent.

Issue #247, opened 2015-11-12
BindableCollection dispatches operations on the collection to the UI thread - why not just the change events?


cdanne:
Hi there,

we have a scenario where a BindableCollection is not only modified by user interaction (drag & drop of list items), but also may get many updates in short time from a background operation. In the latter case, performance is an issue and we set IsNotifying=false before the algorithm starts and re-enable change events afterwards.

That improved performance because it removed UI updates, but didn't make it great... We profiled the app and realized that a major performance killer is the fact that all operations on the collection (Add, Remove etc.) are dispatched to the UI thread via Execute.OnUIThread(...) in the BindableCollection implementation.

Question: Why so? Isn't it enough to have the overwritten OnCollectionChanged() and OnPropertyChanged() raise the **change event** on the UI thread? 
So there would be no waiting for the UI thread to modify the collections in the first place. Dispatching to the UI thread only happens for the change events and that only in case IsNotifying is true.

We did such an implementation and it performs better and seems to work OK in our scenario. Are there any pitfalls with that approach which we oversee or would such an implementation be equally "correct" and faster in such scenarios?

nigel-sampson:
This code predates my taking over the project so I can't speak to the logic of why almost all the operations were wrapped in `Execute.OnUIThead`. On review it does look overzealous.

I'd love to see a pull request with your implementation so we can discuss it further.

On a side note the implementation of `XamlPlatformProvider.OnUIThread` checks for access before using the dispatcher. So if you have a method on a background thread doing a lot of small updates then shifting the whole method over should give you some performance improvements as well.

rohits79:
As a work around: Reactive Extension can be used here to throttle or buffer updates

jongleur1983:
does anyone have any idea how to write tests for this change?
Took the code we have at office (as @cdanne proposed) and changed the BindableCollection accordingly.
All existing caliburn unit tests still running, tested in real-live code at work, but not covered by test code yet. 
Is there any way to write unit tests that properly test thread dispatching (on UI thread, ideally without UI)?

nigel-sampson:
You could potentially test this by creating a fake platform provider that recorded calls to dispatch on the UI thread. 

Your test could then do a number of operations and then assert the amount of ui thread dispatches was correct.

jongleur1983:
I guess the fix up to here doesn't solve the problem but just flips it around:
The original Caliburn implementation did any manipulation of the BindableCollection on the UI-Thread.
[...]

PR #407, opened 2017-01-26
Optionally UI Thread property change marshalling

RafaelSalguero:
When combining property change notification with reactive observables there are situations where is desirable that the property change event is leaved as-is, on its original thread.

Since WPF already does UI Thread marshaling on bindings, would be useful to be able to turn off this behavior.

### DIFF
--- a/src/Caliburn.Micro/PropertyChangedBase.cs
+++ b/src/Caliburn.Micro/PropertyChangedBase.cs
@@ -10,6 +10,11 @@
     [DataContract]
     public class PropertyChangedBase : INotifyPropertyChangedEx {
         /// <summary>
+        /// True to marshal NotifyOfPropertyChange calls to the UI thread
+        /// </summary>
+        public static bool MarshalOnUIThread = true;
+        
+        /// <summary>
         /// Creates an instance of <see cref = "PropertyChangedBase" />.
         /// </summary>
         public PropertyChangedBase() {
@@ -44,7 +49,10 @@
         public virtual void NotifyOfPropertyChange([System.Runtime.CompilerServices.CallerMemberName] string propertyName = null) {
 #endif
             if (IsNotifying && PropertyChanged != null) {
-                Execute.OnUIThread(() => OnPropertyChanged(new PropertyChangedEventArgs(propertyName)));
+                if(MarshalOnUIThread)
+                    Execute.OnUIThread(() => OnPropertyChanged(new PropertyChangedEventArgs(propertyName)));
+                else
+                    OnPropertyChanged(new PropertyChangedEventArgs(propertyName));
             }
         }
 


superware:
Thanks @RafaelSalguero.

Hi @nigel-sampson, can you please consider merging?

We've discussed this a few months ago, but I'm still having issues, mainly with deadlocking on dispose:

```
public class MyViewModel : PropertyChangedBase, IDisposable
{
	private readonly object _sync = new object();
	private readonly IMyService _myService;
	private IMyEntity _myEntity;
	private string _myProperty;
	private string _myProperty2;

	public MyViewModel(IMyService myService)
	{
		_myService = myService;
		_myService.EntityAdded += MyService_EntityAdded;
	}

	private void MyService_EntityAdded(object sender, EntityAddedEventArgs e)
	{
		lock (_sync)
		{
			if (_myEntity != null)
			{
				_myEntity.SomethingChanged -= MyEntity_SomethingChanged;
			}

			_myEntity = e.MyEntity;

			if (_myEntity != null)
			{
				_myEntity.SomethingChanged += MyEntity_SomethingChanged;
			}
		}
	}

	private void MyEntity_SomethingChanged(object sender, EntitySomethingChangedEventArgs e)
	{
		lock (_sync)
		{
			MyProperty = e.MyProperty;
			MyProperty2 = e.MyProperty2; // <-- a working thread is stuck here, waiting for the UI thread which is now disposing
		}
	}

	public string MyProperty
	{
		get { return _myProperty; }
		private set { _myProperty = value; NotifyOfPropertyChange(); }
	}

	public string MyProperty2
	{
		get { return _myProperty2; }
		private set { _myProperty2 = value; NotifyOfPropertyChange(); }
	}

	public void Dispose()
	{
		lock (_sync) // <-- the UI thread is stuck here, waiting for Execute.OnUIThread
		{
			if (_myEntity != null)
			{
				_myEntity.SomethingChanged -= MyEntity_SomethingChanged;
			}

			_myService.EntityAdded -= MyService_EntityAdded;
		}
	}
}
```

Thank you in advance.

nigel-sampson:
Sorry this one slipped through my radar, I've no objection to adding this sort of functionality to the framework.

One thing though is that I'm not sure adding that static to `PropertyChangedBase` is the best place. I'm always cautious when adding new members to the commonly used class in the framework.

I would consider adding it to `IPlatformProvider` since it's static and then be configured solution wide.

I would then add the same behaviour to `BindableCollection` for consistency.

superware:
You simply rock! thanks.

superware:
@RafaelSalguero Can you please revise this PR per @nigel-sampson recommendations? Thanks!

superware:
Hi @nigel-sampson,

Can you please verify the issue for the code snippet I posted earlier? Am I missing something?

Thanks!

nigel-sampson:
Which thread is the Dispose being called on?

nigel-sampson:
Going to close this off with the discussion on #247 on how we'll proceed.

superware:
Hi @nigel-sampson,

Dispose is being called on the UI thread (application shutdown), see comments in the code.

Thanks!

nigel-sampson:
Ah, I'd be really hesitant to use locks on the UI thread for fear of holding stuff up.

superware:
Once the application is being terminated, the IoC container is disposing all classes that implement IDisposable automatically... What would you put inside the VM's Dispose in my code example?

superware:
I'm still having difficulties with PropertyChangedBase (and all inheriting classes) marshalling PropertyChanged to the UI thread.

For example, the application is being disposed on shutdown, if one of the service classes has a thread Join being called while that thread is changing a VM property, then Join will never complete and the application will never exit.

Why not simply add IsPropertyChangedUIMarshallingDisabled (default is false) in IPlatformProvider? isn't this MUCH better than implementing an IPlatformProvider where OnUIThread isn't actually executing something on the UI thread?

I wish there'll be a global way to disable this marshalling behaviour, which is completely unnecessary in WPF.

nigel-sampson:
I'll have another look into this, thanks.

superware:
Hi, did you get a chance to look into this?

superware:
That's great news, I'll gladly test the PR.

nigel-sampson:
Not too rude at all, to be honest it's taking longer than expected mostly because I haven't had a lot of time spare to work on Caliburn.Micro in the last few months beyond triaging issues.

I'm not sure when that will change. Hopefully soon but I can't make any promises.

superware:
@nigel-sampson Hey, as far as I understand this can be a simple non-breaking option, how do you see this? can I help with the implementation? Btw, couldn't find it on the 4.0.0 list :)

nigel-sampson:
I've re-opened this as a reminder.

superware:
Hey @nigel-sampson, any news regarding this? :|

How are you planning to implement this? Via IPlatformProvider? I'm eager to test it..

nigel-sampson:
This has now been included on master and available through the pre-release feed.

superware:
Hi @nigel-sampson, can you please release a new alpha?

nigel-sampson:
Will see what I can do, any reason you can't use the MyGet feed?

superware:
I prefer to keep the NuGet chain of reference. Anyway, is there a time table for the next alpha/beta/release? Thanks.

nigel-sampson:
[`4.0.105-alpha`](https://www.nuget.org/packages/Caliburn.Micro/4.0.105-alpha) has just been published.

superware:
Hi @nigel-sampson, how are you?

It seems PropertyChangeNotificationsOnUIThread isn't being checked on ActionMessage's [PrepareContext](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Caliburn.Micro.Platform/ActionMessage.cs#L455) "Guard Names" mechanism.

The only workaround I found is

`ActionMessage.BuildPossibleGuardNames = m => { return Array.Empty<string>(); };`

On bootstrapper ctor, but it's a hack in order to disable that feature.

nigel-sampson:
Thanks, will look into this.

nigel-sampson:
Can you expand on what problem you're seeing here @superware ?

superware:
ActionMessage is subscribing to PropertyChanged, and the handler is calling Execute.OnUIThread without first checking the platform provider's PropertyChangeNotificationsOnUIThread. If it's false it should update the availability immediately. In my case I'm getting a deadlock, which PropertyChangeNotificationsOnUIThread was designed to prevent on WPF.

nigel-sampson:
Ok, now I get you, it feels weird to be checking `PropertyChangeNotificationsOnUIThread` inside `ActionMessage` as it's not what the property is really representing. Given that the property handler there is updating the UI thread it makes sense that it should use `Execute.OnUIThread`.

The whole thing smells like a problem with the approach as we shouldn't need to guard every call on `Execute.OnUIThread` against this property.

However `Execute.OnUIThread` for WPF should be checking if its on the UI thread and not doing any dispatching if it is already.  Not sure how that would deadlock.

KasperSK:
@nigel-sampson I am closing this as it seems that the issue it represents have been solved at least partly. If you feel that there is any outstanding work in this PR please let us know.

PR #454, opened 2017-06-19
Change the default view creation through DI and/or by directly calling Activator.CreateInstance

TheBigRic:
These changes change the manner in how C.M. creates view as discussed in #339.

The defaults are open for discussing. 

I chose the defaults in how I would like the framework to react. I can't determine this for other users although this implementation is much more towards 'fail fast' as we see many frameworks do nowadays.

Also added unittests for net45.

Although the project builds successfully for some reason it keeps complaining in the IDE that it is missing references to `System` when I select .netstandard1.4. So I can't really see if this code (for x.f.) really will work. We'll have to test that in a x.f. application I think.

Let me know what you think.

### DIFF
--- a/src/Caliburn.Micro.Platform/ViewCantBeCreatedException.cs
+++ b/src/Caliburn.Micro.Platform/ViewCantBeCreatedException.cs
@@ -1,0 +1,8 @@
+﻿namespace Caliburn.Micro {
+using System;
+
+    public class ViewCantBeCreatedException: Exception {
+        public ViewCantBeCreatedException(string message) : base(message) {
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


nigel-sampson:
Thanks for this, sorry I haven't had a chance to dig through it more yet. This lines up a lot with what we've talked about. just not yet sure whether adding this as a configurable option versus leaving the extension point open is the way to go.

TheBigRic:
Currently is does both which gives maybe to much and to hard to document flexibility, I agree. 

What could be changed is letting the extension point open, remove the configuration options and provide the three of methods as statics. The user can in that case use one of the three possibilities (still an unfancy way of configuration) or override the extension point completely.

Issue #470, opened 2017-08-14
SimpleContainer can inject lists through property injection.


Stannieman:
It would be very nice to have list injection with property injection.

EDIT: PR created.

nigel-sampson:
I've addressed this in #471 in order to keep discussion in one place.

Issue #469, opened 2017-08-14
SimpleContainer can use property injection by default


Stannieman:
Property injection could be enabled by default with configuration, which removes the need to explicitly call BuildUp.

EDIT: PR created.

nigel-sampson:
I've addressed this in #471 in order to keep discussion in one place.

Issue #468, opened 2017-08-14
SimpleContainer property injection can be recursive


Stannieman:
BuildUp can call BuildUp again for every injected property to build up the whole graph instead of only the instance passed into it.

EDIT: PR created.

nigel-sampson:
I've addressed this in #471 in order to keep discussion in one place.

Issue #467, opened 2017-08-13
SimpleContainer resolution handler can see the parent type


Stannieman:
When resolving a type TParent that has a dependency TDep then it can be usefull that the handler for TDep knows the type TParent.

For example if you are injecting a logger. The handler for Logger can use TParent to set the class name for the logger.

nigel-sampson:
I've addressed this in #471 in order to keep discussion in one place.

PR #471, opened 2017-08-15
468-469-470 Extend container functionality.

Stannieman:
This are the 3 last issues created by me.

### DIFF
--- a/src/Caliburn.Micro.Tests.NET45/Caliburn.Micro.Tests.NET45.csproj
+++ b/src/Caliburn.Micro.Tests.NET45/Caliburn.Micro.Tests.NET45.csproj
@@ -64,6 +64,7 @@
     <Compile Include="ScopeNamingRouteTests.cs" />
     <Compile Include="ConductorWithCollectionOneActiveTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="SimpleContainerTestModels.cs" />
     <Compile Include="SimpleContainerTests.cs" />
     <Compile Include="StringSplitterTests.cs" />
     <Compile Include="ViewLocatorTests.cs" />
--- a/src/Caliburn.Micro.Tests.NET45/SimpleContainerTestModels.cs
+++ b/src/Caliburn.Micro.Tests.NET45/SimpleContainerTestModels.cs
@@ -1,0 +1,31 @@
+﻿using System.Collections.Generic;
+
+namespace Caliburn.Micro.Tests.NET45
+{
+    public interface IComponent { }
+    public interface IDependency1 { }
+    public interface IDependency2 { }
+    public interface IEnumerableDependency { }
+
+    public class Component : IComponent {
+        public IDependency1 Dependency1 { get; set; }
+        public NonInterfaceDependency NonInterfaceDependency { get; set; }
+    }
+
+    public class Dependency1 : IDependency1 {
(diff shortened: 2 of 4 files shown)


nigel-sampson:
Thanks for the PR, apologies for the delay in looking at all of this. I haven't looked too closely yet at the code, just the discussions.

Just to frame the conversation, I'm really hesitant to extend `SimpleContainer` too far, for a couple of reasons.

1. It stops becoming simple 😄 
2. A fully featured dependency injection container is a whole OSS project in itself and not something I want to do. Better to direct people to fully featured ones if they need something more powerful.

Having said that...

#467 This feels like an feature for a more fully featured framework and not for `SimpleContainer`.

#468 If I'm not mistaken this would be resolved by #469 if `EnablePropertyInjection` was true?

#469 Not sure about this one, it's definitely a breaking change, and I like that you've added it as a setting that's off by default.

#470 Appears to be an oversight given it will be injected as a constructor parameter so we should be consistent between the two. 

#472 Absolutely agree the change of singleton behavior would be a breaking change, it should probably be in it's own PR so other things can be merged without this holding it up. The points around a possible `RequiredDependency` and `DependencyKey` definitely feel like we're straying back to full on container land. Should be adding #374 as part of the next major version.

nigel-sampson:
Thinking about this some more, rather than splitting into separate PR's we could shift it to the `dev/4.0.0` branch.

Stannieman:
The RequiredDependency thing might be overkill actually when I look at it now. This is not really a "feature-feature". If a dependency annotated with it cannot be resolved (which the attribute tries to prevent) then this is actually a programming error. I don't think trying to fix this kind of errors is the container's job.
So let's forget about this feature :)

Then for the other features I proposed: The main reason is that I found the SimpleContainer to be just a little too simple. I started on a new application that is not complex at all, and still I found the current container to be too limited. And if it doesn't meet the requirements for an app of basic to mid complexity then I wonder how many use cases there are left for the current container.
I think adding these few features would make the container useful for a lot more projects.
- Property injection without BuildUp to not have constructors with 100 parameters.
- Recursive property injection to help with the feature above.
- Seeing the target type: example with logger.
- DependencyKey: SimpleContainer already has it but this is just to resolve contructor parameters/properties using the key also.

These things are currently blockers for me and the only reason I need to use a fully featured container.
I have a new container ready (need a few days to properly test it), then I will create a PR for it and you can choose to keep it in v4 or not.

Stannieman:
@nigel-sampson I put together some container that is relatively simple but still suits all my needs and has the features of my previous tickets.
It's up [here](https://github.com/Stannieman/DI/). You can have a look and IF you want it then I'll create a PR for Caliburn 4 + add more testing, if not then I'll release it as a standalone package.

nigel-sampson:
Thanks, I'll take a look but unlikely to switch containers at the point mainly due to the making a release with a lot of breaking changes even worse.

xmedeko:
IMO the code of `BuildUp` in this PR should be merged to master, it does not break BC.
Further more, I recommend not to check `CanWrite && CanWrrite`, but to check: `property.SetMethod != null && !property.SetMethod.IsStatic`.

nigel-sampson:
While the PR's can't be merged at the moment due to the changes in the 4.0.0 codebase, I plan to incorporate parts of these into the 4.0.0 release at some point.

nigel-sampson:
Thanks for this PR, I've incorporated a lot from it.

Issue #462, opened 2017-07-27
Set title of window in xaml


tziemek:
Hi,
we are using CM with WPF and wondered why setting the title on a window in xaml has no effect. I found the responsible code in https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Caliburn.Micro.Platform/net40/WindowManager.cs in the CreateWindow method. There is a check for the IHaveDisplayName interface and existing bindings.

My suggestion is to also do nothing if there is already a title defined, hence i don't have to set the DisplayName property in my viewmodel.

Thanks

nigel-sampson:
That makes sense, happy to take a PR for this. 😄

nigel-sampson:
Resolved by #476 .

Thanks!

PR #476, opened 2017-09-08
Use explicit title for windows and pages

tziemek:
As discussed in #462 the window title is taken from the window property instead of always using a binding.


nigel-sampson:
Looks good to me, thanks for the PR and congrats for your first one!

tziemek:
Glad to help. Didn't really have the time until now.

PR #482, opened 2017-10-11
[Documentation] Fixed spelling of 'Inidicates' with 'Indicates'

llifon:
When using a doc generator (such as Ghostdoc) alongside StyleCop, it would import the spelling mistake and Stylecop would keep flagging it up. This would always happen when overriding OnDeactivate.

A trivial thing but it somehow bothered me greatly enough to do this.


nigel-sampson:
Thanks

PR #485, opened 2017-10-20
Update Xamarin.Forms NavigationPageAdapter

bryanbcook:
This is a small tweak to the Xamarin.Forms NavigationPageAdapter to allow users to apply additional logic when a page is pushed or popped from the navigation stack.

As an example, I have a cross-cutting concern for network connectivity in my application. Users are able to navigate while offline, but I need a visual indicator on each page to show connectivity. Rather than have all pages in the application query the network status on activation, it would be ideal if the network status was provided to them externally as they navigated. With this PR I can subclass NavigationPageAdapter to provide status when pages are pushed/popped.


nigel-sampson:
A good change, thanks.

PR #533, opened 2018-07-12
using nuget package insteed of Expression Blend SDK file reference to System.Windows.Interactivity.dll

bqstony:
for clear dependency and the missing package on a clean VS2017 dev client


nigel-sampson:
I've seen that package before and am a little concerned about it since it's not official source.

bqstony:
I know it is not official but Microsoft gives not a better Solution. And the Download count of 263'960 ist impresive.

bqstony:
@nigel-sampson you have add the dll to the package. Well, this PR is not needed anymorey.

PR #578, opened 2019-01-16
Allow KeyTrigger to match Alt+Key combinations

babackman:
KeyTrigger class in KeyBinding sample could not match Alt-key combinations because the actual key code gets pushed into the SystemKey field.  Fixed KeyTrigger and added to sample to test it.


nigel-sampson:
Thanks.

Issue #593, opened 2019-03-25
IsInitialized and IsActive are set too early


mgnsm:
`IsInitialized `shouldn't be set to `true `*before* the `OnInitializeAsync `method has returned and the view model has actually been initialized. 

Consider a view model that performs some asynchronous initialization:

    public class SampleViewModel : Screen
    {
        protected override async Task OnInitializeAsync(CancellationToken cancellationToken)
        {
            await Task.Delay(5000); //simulate doing something...
        }
    }

And a view that displays a spinner or message to indicate that the view model is being initialized:

    <TextBlock>
        <TextBlock.Style>
            <Style TargetType="TextBlock">
                <Setter Property="Text" Value="Ready!" />
                <Style.Triggers>
                    <DataTrigger Binding="{Binding IsInitialized}" Value="False">
                        <Setter Property="Text" Value="Initializing..." />
                    </DataTrigger>
                </Style.Triggers>
            </Style>
        </TextBlock.Style>
    </TextBlock>

In version 4.0.52-alpha, the "Initializing..." message will never displayed because the `IsInitialized ` property is set to `true `before the `OnInitializeAsync `method is awaited in the `Screen `class.

The same thing applies to `IsActive` and `OnActivateAsync`/`OnDeactivateAsync`.
 
These properties should be set once the tasks have completed. I'll submit another pull request.

a44281071:
Shall we change it to Activating/Activated？Or before/after？
We can test this sample code SampleViewModel => SampleView : Window. Window. Will the window immediately display?

PR #594, opened 2019-03-25
#593 Set the IsInitialized and IsActive properties to true only after…

mgnsm:
… the screen has actually been initialized and activated respectively.


nigel-sampson:
Thanks for work in testing our these interim builds.

PR #596, opened 2019-04-04
Fixes InvalidOperationException in WindowManager

markgould:
This is related to the issue I reported here: https://github.com/Caliburn-Micro/Caliburn.Micro/commit/772ed0bc7b0856973482cf3e5f4218e04a4c5b9c#r33040965

If CanCloseAsync executes too quickly and resumes on the same thread, then calling view.Close will result in an InvalidOperationException. The fix is to only call view.Close again if the DialogResult differs from the cachedDialogResult. If it does not, then we can simply set e.Cancel to false.


nigel-sampson:
Tested this locally and it's looking good. Thanks.

PR #604, opened 2019-05-22
BindableCollection AddRange/RemoveRange NewItems/OldItems are now set

Yinimi:
When adding objects to a BindableCollection the e.NewItems in the CollectionChanged event used to be empty.

Current version:
image

Fixed version:
image

Code snipped:
[Program.txt](https://github.com/Caliburn-Micro/Caliburn.Micro/files/3206761/Program.txt)

### DIFF
--- a/src/Caliburn.Micro.Core/SimpleContainer.cs
+++ b/src/Caliburn.Micro.Core/SimpleContainer.cs
@@ -15,7 +15,7 @@ namespace Caliburn.Micro
         private static readonly Type enumerableType = typeof(IEnumerable);
         private static readonly TypeInfo enumerableTypeInfo = enumerableType.GetTypeInfo();
         private static readonly TypeInfo delegateTypeInfo = delegateType.GetTypeInfo();
-        private Type simpleContainerType = typeof(SimpleContainer);
+        private readonly Type simpleContainerType = typeof(SimpleContainer);
         private readonly List<ContainerEntry> entries;
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


nigel-sampson:
Looks good, thanks, just need to do some legacy research. This is one of those changes that makes me wonder why it hasn't been done before.

tibel:
be aware of https://github.com/dotnet/wpf/issues/1887

Phyxion:
Just in case it might be useful: some collections (such as `ListCollectionView`) do not support ranged actions and throw. It works with `Reset`.

Yinimi:
I rebased my fork and updated the code

vb2ae:
@Yinimi Code looks good.  Need to do some testing this weekend

vb2ae:
@KasperSK it is testing out ok.   What do you think about merging it in?

KasperSK:
@vb2ae Could we test the case mentioned in [dotnet/wpf#1887](https://github.com/dotnet/wpf/issues/1887)? Seems bad if we could get runtime exceptions when running WPF.

vb2ae:
Tested with this code

        public void btnChangeCollection()
        {
            Random rnd = new Random();
            var lst = new List<int>();
            for (int x = 0; x < 100; x++)
            {
                lst.Add(Numbers[x]);
            }

            Numbers.RemoveRange(lst);
            var lstToAdd = new List<int>();
            for (int z = 0; z < 100; z++)
            {
                lstToAdd.Add(rnd.Next(200000));
            }

            Numbers.AddRange(lstToAdd);
        }

Numbers was bound to a listbox.  AddRange and RemoveRange throw exceptions like @tibel  mentioned

image

KasperSK:
Looks like we should not merge this before WPF has fixed the issue then.

PR #625, opened 2019-08-01
Set WindowConductor as protected

brunomoreirafaro:


nigel-sampson:
Thanks for submitting a PR. Would be a good idea to make some of the methods virtual as well to make it easier to extend?

PR #609, opened 2019-05-30
Add Dot Net Core 3.0 Support 

Rmurray0809:
Added a new platform for .net Core (literally copy and pasted net46 folder for now).  This breaks builds targeting UAP.   But assuming MSBuild gets fixed to not confuse the libraries UAP needs this could then be merged.

PR #626, opened 2019-08-09
Add .NET Core 3.0 target to Caliburn.Micro.Platform

mjuen:
As #609 seems to be stuck at the moment I took the liberty to replicate the changes and include the review suggestions from @nigel-sampson.

I also renamed the folder containing the shared files from `net46` to `net46-netcore`.

### DIFF
--- a/src/Caliburn.Micro.Platform/ActionMessage.cs
+++ b/src/Caliburn.Micro.Platform/ActionMessage.cs
@@ -191,7 +191,7 @@
             }
             else currentElement = context.View;
 
-#if NET
+#if NET || NETCOREAPP
             var binding = new Binding {
                 Path = new PropertyPath(Message.HandlerProperty), 
                 Source = currentElement
--- a/src/Caliburn.Micro.Platform/ActionMessage.cs
+++ b/src/Caliburn.Micro.Platform/ActionMessage.cs
@@ -191,7 +191,7 @@
             }
             else currentElement = context.View;
 
-#if NET || NETCOREAPP
+#if NET || NETCORE
             var binding = new Binding {
                 Path = new PropertyPath(Message.HandlerProperty), 
                 Source = currentElement
--- a/src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.csproj
+++ b/src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.csproj
@@ -7,6 +7,11 @@
     <RootNamespace>Caliburn.Micro</RootNamespace>
   </PropertyGroup>
 
+  <!-- Workaround for https://github.com/onovotny/MSBuildSdkExtras/issues/173 -->
+  <PropertyGroup Label="Android DesignTimeBuild error workaround">
+    <DesignTimeBuild>false</DesignTimeBuild>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'Xamarin.iOS10' or '$(TargetFramework)' == 'MonoAndroid80'">
     <DefineConstants>XFORMS</DefineConstants>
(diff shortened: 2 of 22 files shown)


src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.csproj
```diff
@@ -74,7 +80,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Remove="Platforms\net46\AssemblyInfo.cs" />
+    <None Remove="Platforms\net46-netcore\AssemblyInfo.cs" />
```
nigel-sampson: Not related to this PR, but I suspect the whole `ItemGroup` can be removed. VS tends to screw up projects when removing files that are wildcard included.

src/Caliburn.Micro.Platform/ActionMessage.cs
```diff
@@ -191,7 +191,7 @@ void ElementUnloaded(object sender, RoutedEventArgs e)
             }
             else currentElement = context.View;
 
-#if NET
+#if NET || NETCOREAPP
```
nigel-sampson: Where is `NETCOREAPP` being defined? Would expect to see something in [`Directory.Build.targets`](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Directory.Build.targets).

nigel-sampson:
Looking really good, thanks. Need to sort out why PR builds aren't happening to test this out.

src/Caliburn.Micro.Platform/ActionMessage.cs
mjuen: `NETCOREAPP` is a predefined symbol for .NET Core as target framework.
https://docs.microsoft.com/en-us/dotnet/standard/frameworks

Sorry, I did not check the `Directory.Build.targets` file. I have replaced the `NETCOREAPP` symbol with `NETCORE`, which is now defined in [Directory.Build.targets](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/df87adaac6208f07ea8fd727e5c2a2eed75512de/src/Directory.Build.targets). (df87adaac6208f07ea8fd727e5c2a2eed75512de)

src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.csproj
mjuen: Removed the ItemGroup in df87adaac6208f07ea8fd727e5c2a2eed75512de

nigel-sampson:
I've enabled PR builds on forks so any further changes should be built for you.

mjuen:
> 
> 
> I've enabled PR builds on forks so any further changes should be built for you (in theory).

Could you change the SDK to one that supports .NET Core 3.0?

Build is currently failing with `The current .NET SDK does not support targeting .NET Core 3.0.  Either target .NET Core 2.2 or lower, or use a version of the .NET SDK that supports .NET Core 3.0`.

nigel-sampson:
Having some issues updating the pipeline. 
> ##[error]C:\Users\VssAdministrator\.nuget\packages\msbuild.sdk.extras\2.0.31\Sdk\Sdk.props(16,5): Error MSB4184: The expression "[System.IO.Path]::GetDirectoryName('')" cannot be evaluated. The path is not of a legal form.

Will work through it some more tonight.

nigel-sampson:
Glad we managed to get there together

nigel-sampson:
Thanks for the hard work @mjuen

mjuen:
No problem :)

It is truly nerve wracking when you assume that the pipeline now works after fixing several errors and then the next one appears.

nigel-sampson:
Hate having to fix pipelines, leaves a lot of terrible commit history.

PR #633, opened 2019-09-06
Rename methods with "Aysnc" to "Async"

louistio:
There are two typos where a method suffix is "Aysnc" instead of "Async":
- The private method CanCloseAysnc of NavigationPageAdapter.
- The public method InitialiseAysnc of private class WindowConductor.


nigel-sampson:
Thanks, this is a fantastic example of "semantic exhaustion".

PR #637, opened 2019-09-24
Unset Parent on Conductors' Items.Clear()

### LINKED ISSUE #465: Conductor<T>.Collection.OneActive/AllActive - IChild.Parent not reset on clear
```csharp
void Main()
{
	var conductor = new Conductor<TestItem>.Collection.AllActive();
	
	var t1 = new TestItem();
	var t2 = new TestItem();
	var t3 = new TestItem();
	conductor.Items.Add(t1);
	conductor.Items.Add(t2);
	conductor.Items.Add(t3);
	if (t1.Parent == null || t2.Parent == null || t3.Parent == null)
		throw new Exception("t1.Parent + t2.Parent + t3.Parent should have been set by conductor");
	conductor.Items[0] = null;
	if (t1.Parent != null)
		throw new Exception("t1.Parent should have been set to <null> by conductor");
	conductor.Items.Remove(t2);
	if (t2.Parent != null) 
		throw new Exception("t2.Parent should have been set to <null> by conductor");
	conductor.Items.Clear();
	if (t3.Parent != null)
		throw new Exception("t3.Parent should have been set to <null> by conductor");
}
```

The code above fails on 't3.Parent != null', it seems like the conductor implementation only resets the child's parent property on a normal or index-based removal, but not on a clear operation.

adamgauthier:
ObservableCollection<T> raises a reset notification when Clear() is
called, but NotifyCollectionChangedEventArgs.OldItems is not populated.
BindableCollection<T>.RemoveRange has a similar issue because it
suppresses individual remove notifications and raises a reset
notification.

Because of this, Caliburn conductors can't reliably unset the Parent
property on removed IChild instances after bulk removals.

This adds a CollectionItemsRemoved event to BindableCollection<T> for
bulk removals represented by reset notifications. Conductors use it to
clear IChild.Parent after Clear() and RemoveRange(), while single-item
removals continue to use the standard CollectionChanged notification.

Fixes #465


louistio:
More discussion in #465 and on [stackoverflow](https://stackoverflow.com/questions/224155/when-clearing-an-observablecollection-there-are-no-items-in-e-olditems).

louistio:
```
Caliburn.Micro.Core.Tests.EventAggregatorSubscribing.Caliburn.Micro.Core.Tests.EventAggregatorSubscribing.A_valid_subscriber_is_assigned_as_a_handler_its_message_type
Assert.True() Failure
Expected: True
Actual:   False
```

Not sure why that test is failing, everything is green locally. 🤔

vb2ae (changes requested):
Thanks for merging in main.   Could you update the version number in version.json to 6.0-beta.  

{
    "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
    "version": "6.0-beta",
    "nugetPackageVersion": {
        "semVer": 2
    },
    "publicReleaseRefSpec": [
        "^refs/heads/master$"
    ],
    "cloudBuild": {
        "buildNumber": {
            "enabled": true,
            "includeCommitId": {
                "when": "nonPublicReleaseOnly",
                "where": "buildMetadata"
            }
        },
    "setVersionVariables": true
    }
}

adamgauthier:
@vb2ae Sure, done!

vb2ae (approved):
Thank you for your pr.   I approve the changes.  I am wondering if we should add a bool property to IChild LegacyMode which will stop setting the parent to null if true. (allow for opt it if breaks their app)

adamgauthier:
Interesting idea. Honestly, I feel like it's unlikely apps intentionally rely on the old behavior. It was inconsistent with the existing single-item remove/replace semantics, where Parent is already cleared, and the stale Parent reference seems more like an accidental bug.

PR #639, opened 2019-09-30
Upgrade MSBuild.Sdk.Extras to 2.0.54

louistio:


nigel-sampson:
Is this intended to resolve the current build failures?

louistio:
Well it was an attempt at least 😅 Seems like there are other problems

louistio:
The actual build seems to pass with this change, but there is a problem with `samples\features\Features.Forms\Features.Forms.Android\Resources\Resource.Designer.cs`. I didn't know these sample projects existed. 🤐

nigel-sampson:
Yeah the `features` samples (but not the `setup`) projects build from the build output as a simple integration test.

nigel-sampson:
Thanks for this even though it wasn't merged. 

Still need to dig into the features build issue.

louistio:
No worries, I wish I could help more, I don't even have Xamarin dev tools installed and I've never done any before so I wouldn't even know where to start.

PR #641, opened 2019-10-06
Remove MSBuildSdkExtras workarounds

gpetrou:
It looks like https://github.com/onovotny/MSBuildSdkExtras/issues/173 and https://github.com/onovotny/MSBuildSdkExtras/issues/174 have been resolved in newer MSBuildSdkExtras version.

### DIFF
--- a/azure-pipeline.yml
+++ b/azure-pipeline.yml
@@ -32,13 +32,6 @@ steps:
     msbuildArgs: '/t:Restore'
     configuration: '$(buildConfiguration)'
 
-- task: PowerShell@2
-# Workaround for https://github.com/onovotny/MSBuildSdkExtras/issues/174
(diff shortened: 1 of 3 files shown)


nigel-sampson:
Well it doesn't look to be making the build worse, still looking into the issue.  Not sure why we would see a `Resource` class in `Caliburn.Micro.Core` or `Caliburn.Micro.Platform` as neither have any resources. 

Looking at `src\Caliburn.Micro.Core\obj\Release\monoandroid80\Resource.designer.cs` shows an empty file on my machine.

I also can't see anything in the [features `Resource.designer.cs`](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/samples/features/Features.Forms/Features.Forms.Android/Resources/Resource.designer.cs) that's referencing this other class.

PR #674, opened 2020-02-28
added default CancellationTokens to common methods (fixes #663)

### LINKED ISSUE #663: Default CancellationToken for async methods
I find myself passing default(CancellationToken) all over, although most of MS async implementations support a default CancellationToken parameter. Any thoughts?

Mrxx99:
This adds default CancellationTokens to many common methods that are likely to be called by the user of this library (I did not add it to the event like methods like OnInitalized).
This fixes #663 
Since for some mehtods there was already a overload without a CancellationToken I tried to be consistent inside a class. That means if the class already conatined such overloads for some methods I did add overloads without CancellationToken for the remaining ones in this class, otherwise I provided a default value for the existing declarations.


nigel-sampson:
Thanks for this. Will review when I can.

Issue #700, opened 2020-06-02
FileLoadException when trying to load a .netstandard DLL into a .netframework WPF build


KasperSK:
When I try and reference a .NetStandard dll that references Caliburn.Micro.Core from a WPF program that targets .NetFramework I get the following error:

FileLoadException: Could not load file or assembly 'Caliburn.Micro.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

The exception has a Fusion log trace with it:

WRN: Comparing the assembly name resulted in the mismatch: PUBLIC KEY TOKEN
ERR: Failed to complete setup of assembly (hr = 0x80131040). Probing terminated.

I suspect that it is because the Caliburn.Micro.Core package for the .NetFramework is strong signed, since I can get this to work if i target .NetCore instead of .NetFramework. 

I will upload a test repo later today that illustrates the error.

nigel-sampson:
Thank you.

willson556:
I _think_ I'm running into a related issue. With VS 16.6.x, I'm unable to build a project under `net472` that references a `netstandard2.0` project utilizing `Caliburn.Micro.Core`.

I get the error:
```
AModel.cs(5,27): error CS0012: The type 'PropertyChangedBase' is defined in an assembly that is not referenced. You must add a reference to assembly 'Caliburn.Micro.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=null'. [C:\Users\willsont\projects\CaliburnMicroMultiTarget\WPFLib\WPFLib.csproj]
```

The `netcoreapp3.1` target builds fine. I've put a minimal reproduction project up at https://github.com/willson556/caliburnmicro-multitargeting-issue. Thanks!

nigel-sampson:
Thanks for the reproduction, I can see the same error on my machine and it's pretty perplexing.

nigel-sampson:
Found that changing `Core.csproj` to the following lets it build.
``` xml
<PropertyGroup>
  <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
</PropertyGroup>
```

nigel-sampson:
My guess is because you end up with a conflict of which `Caliburn.Micro.Core` assembly to use, the one built for netstandard vs the one for net472.

willson556:
[...]

PR #704, opened 2020-06-10
Strong-name all Assemblies

willson556:
Possible solution to #700
Tweaks for consistent strong-naming.
Also, reduced the number of targets required.


nigel-sampson:
Reverted my commit, feel free to rebase, let's leave the target frameworks alone.

nigel-sampson:
Although does having the different frameworks solve you issue?

willson556:
Just finished testing and it works for me regardless of whether the frameworks change is present. Rebased as requested and removed the `TargetFrameworks` change in the `Caliburn.Micro.Core` project.

nigel-sampson:
Thanks

PR #719, opened 2020-09-23
update readme

vb2ae:
remove no longer being maintained


nigel-sampson (approved):
Happy to drop this for now and will re-evaluate in a few months.

PR #720, opened 2020-09-23
Fixes the parsing of negative numbers in action parameters #521

### LINKED ISSUE #521: Parsing of negative numbers passed to a Target of an Action does not work
The following XMAL statement is not correctly parsing the '-0.1' value:

`cal:Message.Attach="[Event Click] = [Action TargetTemperatureIncrement(-0.1)]"`

image

The following XMAL statement works correctly.

`cal:Message.Attach="[Event Click] = [Action TargetTemperatureIncrement(0.1)]"`

image

The workaround in this case is to have two methods, one a decrement and one an increment.

KasperSK:
Fixes #521 

Changes the char.IsNumber to decimal.TryParse as the try parse will take the sign into account.

### DIFF
--- a/src/Caliburn.Micro.Platform.Tests/ParserTests.cs
+++ b/src/Caliburn.Micro.Platform.Tests/ParserTests.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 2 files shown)


nigel-sampson (commented):
Looks good to me.

PR #722, opened 2020-09-25
Parameter was not made aware of owner because of assignment order #517

### LINKED ISSUE #517: Guard method not firing on Xamarin.Forms
Xamarin.Forms 2.5.0.280555
Caliburn.Micro.Xamarin.Forms 3.2.0

MainPageViewModel.cs
```csharp
using Caliburn.Micro;

namespace Client
{
    public class MainPageViewModel : Screen
    {
        private int count = 1;
        public int Count
        {
            get { return count; }
            set
            {
                count = value;
                NotifyOfPropertyChange(() => Count);
            }
        }

        public void CountUp(int i)
        {
            Count += i;
        }

        public bool CanCountUp(int i)
        {
            if (i > 10)
                return false;

            if (Count < 10)
                return true;
            else
                return false;
        }
    }
}
```

MainPageView.xaml
```xaml
<?xml version="1.0" encoding="utf-8" ?>
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:local="clr-namespace:AS6_AndroidClient"
             xmlns:cm="clr-namespace:Caliburn.Micro.Xamarin.Forms;assembly=Caliburn.Micro.Platform.Xamarin.Forms"
             Title="TEST"
             x:Class="Client.MainPageView">
    <ContentPage.Content>
        <StackLayout Orientation="Vertical">
            <Label x:Name="Count" Text="{Binding Count}"/>
            <Button Text="{Binding Delta" x:Name="ButtonDelta">
                <Button.Triggers>
                    <EventTrigger Event="Clicked">
                        <cm:ActionMessage MethodName="CountUp" AssociatedObject="{x:Reference ButtonDelta}" >
                            <cm:Parameter Value="{Binding Source={x:Reference Count}, Text}"/>
                        </cm:ActionMessage>
                    </EventTrigger>
                </Button.Triggers>
            </Button>
        </StackLayout>
    </ContentPage.Content>
</ContentPage>
```

KasperSK:
When the OnAttached event was fired the Parameters had not yet been added to the parameters collection and was never made aware of the owning ActionMessage. A CollectionChanged EventHandler on the parameters collection is used to call MakeAwareOf.

Fixes #517

### DIFF
--- a/src/Caliburn.Micro.Platform/Platforms/Xamarin.Forms/ActionMessage.cs
+++ b/src/Caliburn.Micro.Platform/Platforms/Xamarin.Forms/ActionMessage.cs
@@ -39,6 +39,10 @@
         public ActionMessage()
         {
             Parameters = new AttachedCollection<Parameter>();
+            Parameters.CollectionChanged += (s, e) =>
+            {
+                e.NewItems.OfType<Parameter>().Apply(x => x.MakeAwareOf(this));
+            };
         }
 
         /// <summary>


vb2ae (approved):
Change looks good.

PR #725, opened 2020-10-04
Update Caliburn.Micro.Platform.csproj

vb2ae:
move package reference to released version


nigel-sampson (approved):

nigel-sampson:
Looks like the build is failing on this version

vb2ae:
sorry about that undoing change

nigel-sampson:
All good, no need to revert the change as this is only just a PR. it looks like the package version doesn't exist.

Mrxx99:
The whole package was unlisted on nuget.org: https://www.nuget.org/packages/Microsoft.WindowsDesktop.App

nigel-sampson:
Fun, assume there's a better replacement to use instead then?

nigel-sampson:
Should the latest commit be on this branch / PR?

vb2ae:
no removed it

PR #726, opened 2020-10-06
Reinstate XF package to make XF available on more platforms

KasperSK:
As discussed in #724 this pull request will reinstate the Xamarin.Forms package to allow for Xamarin.Forms on UWP or any other platform that Caliburn.Micro and Xamarin.Forms both support. 

Fixes #724

PR #730, opened 2020-11-26
Update .net framework version and .net core version

vb2ae:
Added .net 5 support for windows wpf bug #731 
updated the .net core version from unsupported 3.0 to 3.1 bug #734

updated the ,net framework version to 4.5.2 from unsupported 4.5 in Caliburn.Micro.Core bug #733
updated the ,net framework version to 4.6.1 from 4.5 to match @KasperSK Xamarin forms changes https://github.com/Caliburn-Micro/Caliburn.Micro/pull/726
Update the .net core to use the nuget Microsoft.Windows.Compatibility from Microsoft.WindowsDesktop.App

updated the nuget packages Microsoft.SourceLink.GitHub and Nerdbank.GitVersioning
updated xuint to latest version

updated global.json to use sdk version 5.0.100 so it will compile when .net 5 is installed locally
updated global.json to use MSBuild.Sdk.Extras version to 2.1.2

updated the features apps to build with the new .net core and .net framework versions

Added .net 5 sample to features


src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.csproj
```diff
@@ -76,4 +84,9 @@
     <Compile Include="ViewModelLocator.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="Microsoft.SourceLink.GitHub" Version="1.0.0" />
+    <PackageReference Update="Nerdbank.GitVersioning" Version="3.3.37" />
```
nigel-sampson, line 89: You'll want to update the versions in https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Directory.Build.props which will affect all projects rather that just individual projects.

nigel-sampson (changes requested):

src/Caliburn.Micro.Platform/Caliburn.Micro.Platform.csproj
vb2ae: Updated

nigel-sampson: Should be able to drop the references from the projects as well.

vb2ae: done

nigel-sampson (approved):

Issue #556, opened 2018-11-05
Compatibility inquiry - AvaloniaUI?


MostHated:
Hey there,
    I apologize if this is a silly question, but I am fairly new to all of this. I was wondering if this framework is compatible with the AvaloniaUI cross-platform UI Framework by chance? https://github.com/AvaloniaUI/Avalonia

Thanks,
-MH

mvermef:
I think you will find that there will be a tremendous amount of overlap between both frameworks.  It's gonna come down to which ever suits your needs the best.

MostHated:
Oh, I see. I was under the impression that this framework was to allow easier implementation of MVVM pattern (such as essentially making necessary commands shorter and things of that nature) and AvaloniaUI was just to allow you to create a UI that was cross-platform compatible (Linux, Mac, etc). Was I mistaken on that?

nigel-sampson:
It looks interesting, I'm not sure what the effort would be required to fully implement Caliburn.Micro in Avalonia.

In terms of doing the bare minimum in order get your view models written in Caliburn.Micro to be used in Avalonia is pretty small, if you'd like to spike a prototype I can definitely help you there.

MostHated:
Great news! I will be sure to keep an eye out for that.

Thanks,
-MH

MostHated:
I have not used either yet, I have been mostly researching and watching tutorial videos and just following along on learning all of this stuff (including C#). Maybe one day I might have the skills to manage something like that, lol.

libin85:
@MostHated : I guess currently its just RxUI that supports AvaloniaUI

davidhenley:
@libin85 and Prism

megazyz:
Hello,

I have added initial support for Avalonia. It seems to work fine with some minor shortcomings.
https://github.com/megazyz/Caliburn.Micro.Avalonia/tree/avalonia

How to start:
https://github.com/megazyz/Caliburn.Micro.AvaloniaSample

What do you think about it?

ReactiveUI support is also being preapred, I will commit later.

Stannieman:
@vb2ae I see that there are breaking changes in the upcoming Avalonia version (11) which is in preview now (mainly IAvaloniaObject is removed or replace, I did not check futher). Is that the version we should target? Then I will see if I can make some adjustments to support it.

vb2ae:
[...]

PR #738, opened 2021-01-19
[WIP] AvaloniaUI support

megazyz:
This makes Caliburn.Micro work with [AvaloniaUI](https://avaloniaui.net/).
Related to #556.
Needs review and testing.

### DIFF
--- a/src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/WindowManager.cs
+++ b/src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/WindowManager.cs
@@ -1,0 +1,303 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Avalonia;
+using Avalonia.Controls;
+using Avalonia.Controls.ApplicationLifetimes;
+using Avalonia.Controls.Primitives;
+using Avalonia.Data;
+using Avalonia.Input;
+using Caliburn.Micro;
+using JetBrains.Annotations;
+
+namespace Caliburn.Micro
+{
+    /// <summary>
+    /// A service that manages windows.
+    /// </summary>
+    public class WindowManager : IWindowManager
+    {
+        /// <summary>
+        /// Shows a modal dialog for the specified model.
+        /// </summary>
+        /// <param name="rootModel">The root model.</param>
+        /// <param name="context">The context.</param>
+        /// <param name="settings">The dialog popup settings.</param>
+        /// <returns>The dialog result.</returns>
+        public virtual async Task ShowDialogAsync(object rootModel, object context = null, IDictionary<string, object> settings = null)
+        {
+            var window = await CreateWindowAsync(rootModel, true, context, settings);
+            
+            await window.ShowDialog(InferOwnerOf(window));
+        }
+
+        /// <summary>
+        /// Shows a window for the specified model.
+        /// </summary>
+        /// <param name="rootModel">The root model.</param>
+        /// <param name="context">The context.</param>
+        /// <param name="settings">The optional window settings.</param>
+        public virtual async Task ShowWindowAsync(object rootModel, object context = null, IDictionary<string, object> settings = null)
+        {
+            /*
+            NavigationWindow navWindow = null;
+
+            var application = Application.Current;
+            if (application != null && application.MainWindow != null)
+            {
+                navWindow = application.MainWindow as NavigationWindow;
+            }
+
+            if (navWindow != null)
+            {
+                var window = await CreatePageAsync(rootModel, context, settings);
+                navWindow.Navigate(window);
+            }
+            else
+            {
+            */
+                var window = await CreateWindowAsync(rootModel, false, context, settings);
+
+                window.Show();
+            /*}*/
+        }
+
+        /// <summary>
+        /// Shows a popup at the current mouse position.
+        /// </summary>
+        /// <param name="rootModel">The root model.</param>
+        /// <param name="context">The view context.</param>
+        /// <param name="settings">The optional popup settings.</param>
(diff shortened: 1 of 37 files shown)


KasperSK:
This look interesting, I will try and a review it once I get an hour or two of free time.

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/WindowManager.cs
```diff
@@ -0,0 +1,303 @@
...
+        protected virtual Window EnsureWindow(object model, object view, bool isDialog)
+        {
+
+            if (view is Window window)
+            {
+                var owner = InferOwnerOf(window);
+                if (owner != null && isDialog)
+                {
+                    //TODO: (Avalonia) Set window owner
+                    //window.Owner = owner;
```
vb2ae, line 172: Should this be commented out?

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/WindowManager.cs
```diff
@@ -0,0 +1,303 @@
...
+                };
+
+                window.SetValue(View.IsGeneratedProperty, true);
+
+                var owner = InferOwnerOf(window);
+                if (owner != null)
+                {
+                    window.WindowStartupLocation = WindowStartupLocation.CenterOwner;
+                    //TODO: (Avalonia) Set window owner
+                    //window.Owner = owner;
```
vb2ae, line 190: do we need to set window owner here?

vb2ae:
Could you add a sample AvaloniaUI app to the samples also?

megazyz:
> Could you add a sample AvaloniaUI app to the samples also?

Done.

I'm not sure where it would be ideal to start the CM bootstrapper. Now it's created in App.OnFrameworkInitializationCompleted() in App.axaml.cs. Another solution might be to include it as a resource in App.axaml like in WPF.

KasperSK:
Should we namespace this similarly to the Xamarin.Forms package? I don’t think we can determine whether or not a project uses Avalonia otherwise.

vb2ae:
I agree we need an Avalonia package similar to the xamarin.forms

KasperSK:
Also I think this was made before we reintroduced the `Xamarin.Forms` package so the branch needs to be rebased and the `Caliburn.Micro.Platform.Core` should be referenced as it is for the `Xamarin.Forms` package and the regular `Caliburn.Micro` package.

Stannieman:
What are the steps left to have this into Caliburn? I could help some but am not sure where in the process we are atm.

vb2ae:
will work on fixing merge conflicts and see if we can get it merged in this weekend.  It will need testing

KasperSK:
I tried a while back to include Avalonia in the features sample project, and ran into some issues with the way navigation was handled via frame adapter. I think if we can make it work with the features sample project it will be a good test.

vb2ae:
Created a branch of my own to keep working on this.   https://github.com/Caliburn-Micro/Caliburn.Micro/tree/556-compatibility-inquiry-avaloniaui The biggest issue I have remaining is getting 

          cal:Message.Attach="RegisterFrame($source)" 

events to fire.

Stannieman:
> Created a branch of my own to keep working on this. https://github.com/Caliburn-Micro/Caliburn.Micro/tree/556-compatibility-inquiry-avaloniaui The biggest issue I have remaining is getting
> 
> ```
>       cal:Message.Attach="RegisterFrame($source)" 
> ```
> 
> events to fire.

I have a fix/improvement, shall I make a pull request to your branch or how will we coordinate this?

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/WindowManager.cs
mathmakesmemighty: judging from the methods name, it should. 
why  has nobody answered your review-questions for such a long time?

Stannieman: I think the question is because this flow is not valid for Avalonia.
You pass the owner as parameter to ShowDialog (mandatory) or Show (optional). So it is really a business/architectural decision if we divert from traditional Caliburn way of doing it at the cost of introducing inconsistencies between platforms.

Stannieman: Also beware that work has continued here: https://github.com/Caliburn-Micro/Caliburn.Micro/tree/556-compatibility-inquiry-avaloniaui
It can use a rebase btw cause it is not compatible anymore with current master, there are conflicts due to merge of MAUI support.

danwalmsley:
@megazyz and all! I am from Avalonia! I would like to help resurrect this PR.

@megazyz how complete is this?

danwalmsley:
> @megazyz and all! I am from Avalonia! I would like to help resurrect this PR.
> 
> @megazyz how complete is this?

@vb2ae I have seen your branch, testing it out! Would love to help get it finished. Let me know of how I can assist. For now I shall test it out find my way around!

Stannieman:
The latest Avalonia nightlies have some breaking changes.
One has to do with different parameters in InitializeComponent which is called through reflection by Caliburn. I have a fix for this locally that I'll create a PR for.
One other issue has to do with registering event triggers, not sure exactly what is the cause there.

Then we need to also rebase this branch on master because I am sure there are some conflicts. Some other thing might still be missing that I don't encounter in my project, @vb2ae has more details about them I think.

megazyz:
I apologize for my absence. Thank you for looking into this and continuing the development.

Stannieman:
@megazyz you do this in your free time so no apologies needed mate

PR #739, opened 2021-01-29
Action message select overloaded function #314

### LINKED ISSUE #314: Message.Attach picks wrong method because of similar name
Hello.

I've experienced an issue with this commit: https://github.com/taori/SE-Calculator/commit/9a168ba34b9b422626f054a838ae69a0fa3df1e5 

The error i'm running into is with NewShipViewModel.cs basically i have 2 methods with the same name where only the parameter type is different. However it will only enter the method with Delete(Thruster item) if i click on both types. For an energy source it too will enter the method with a thruster, but pass a null as parameter.

I suppose a little fix is required for caliburn to pick the correct method here?

KasperSK:
This is a proposal on how to fix #314, the implementation in this pull request is more strict than Caliburn.Micro has been in the past.

If there is no method found that matches the parameter types supplied it will return null. This means that a string wont be converted to int even if it is possible. It will take into account derived classes.

I would love some feedback on this @vb2ae, @nigel-sampson.

### DIFF
--- a/src/Caliburn.Micro.Platform/ActionMessage.cs
+++ b/src/Caliburn.Micro.Platform/ActionMessage.cs
@@ -373,62 +373,7 @@
         /// </summary>
         /// <returns>The matching method, if available.</returns>
         public static Func<ActionMessage, object, MethodInfo> GetTargetMethod = (message, target) => {
-#if WINDOWS_UWP
-            //return (from method in target.GetType().GetRuntimeMethods()
-            //        where method.Name == message.MethodName
-            //        let methodParameters = method.GetParameters()
-            //        where message.Parameters.Count == methodParameters.Length
-            //        select method).FirstOrDefault();
-
-            var methods = (from method in target.GetType().GetRuntimeMethods()
-                        where method.Name == message.MethodName
-                        let methodParameters = method.GetParameters()
-                        where message.Parameters.Count == methodParameters.Length && message.Parameters.OfType<Parameter>().Zip<Parameter, ParameterInfo, bool>(methodParameters,
-                        (Parameter parameter, ParameterInfo info) => info.ParameterType.IsInstanceOfType(parameter.Value)).All(b => b)
-                        select method);
-
-            MethodInfo returnMethodInfo = null;
-            foreach (MethodInfo method in methods)
-            {
-                returnMethodInfo = method;
-                if (method.GetParameters().Zip(message.Parameters.OfType<Parameter>(), (info, parameter) =>
-                    parameter.Value.GetType().IsAssignableFrom(info.ParameterType)
-                ).All(b => b))
-                {
-                    break;
-                }
-            }
-
-            return returnMethodInfo;
-
-#else
-            //return (from method in target.GetType().GetMethods()
-            //where method.Name == message.MethodName
-            //let methodParameters = method.GetParameters()
-            //where message.Parameters.Count == methodParameters.Length
-            //select method).FirstOrDefault();
-
-            var methods = (from method in target.GetType().GetMethods()
-                           where method.Name == message.MethodName
-                           let methodParameters = method.GetParameters()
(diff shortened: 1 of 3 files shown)


src/Caliburn.Micro.Platform/ActionMessage.cs
```diff
@@ -374,20 +374,46 @@ void ElementUnloaded(object sender, RoutedEventArgs e)
         /// <returns>The matching method, if available.</returns>
         public static Func<ActionMessage, object, MethodInfo> GetTargetMethod = (message, target) => {
 #if WINDOWS_UWP
-            return (from method in target.GetType().GetRuntimeMethods()
-                    where method.Name == message.MethodName
-                    let methodParameters = method.GetParameters()
-                    where message.Parameters.Count == methodParameters.Length
-                    select method).FirstOrDefault();
+            //return (from method in target.GetType().GetRuntimeMethods()
```
vb2ae: could we remove the commented out code?  If we need to see old code it should be in the history for the repo

src/Caliburn.Micro.Platform/ActionMessage.cs
```diff
@@ -374,20 +374,46 @@ void ElementUnloaded(object sender, RoutedEventArgs e)
...
+                {
+                    break;
+                }
+            }
+
+            return returnMethodInfo;
+
 #else
-            //return 
+            //return (from method in target.GetType().GetMethods()
```
vb2ae: Could we remove commented out code?

src/Caliburn.Micro.Platform/Platforms/Xamarin.Forms/ActionMessage.cs
```diff
@@ -313,11 +313,31 @@ public override string ToString()
         /// <returns>The matching method, if available.</returns>
         public static Func<ActionMessage, object, MethodInfo> GetTargetMethod = (message, target) =>
         {
-            return (from method in target.GetType().GetRuntimeMethods()
-                    where method.Name == message.MethodName
-                    let methodParameters = method.GetParameters()
-                    where message.Parameters.Count == methodParameters.Length
-                    select method).FirstOrDefault();
+            //return (from method in target.GetType().GetRuntimeMethods()
```
vb2ae: Could we remove the commented out code?

vb2ae:
@KasperSK  the code looks good to me.  When the code gets merged to master could you add a blog post about this?

KasperSK:
@vb2ae Sure a blog post would make sense, and I would love to write it.

I had a thought what if we made the strict parsing optional? That way we could still get the old behavior where strings are converted to numbers.

vb2ae:
@KasperSK true a flag to turn it on or off would be a good idea.  Do you think strict parsing should be on or off by default?

nigel-sampson:
Yeah, I suggest making this an opt in feature as it potentially could cause a breaking change in what action gets called.

KasperSK:
@vb2ae I think current behavior should be the default. I was thinking about making a global flag to opt in for all action messages and a local flag to overwrite behavior for a single action message. This will enable the old functionality when needed otherwise I think we need to look into having the Parser assigning type when parsing and I don't think that would be a good idea.

@nigel-sampson Do you think it would be best to put the global flag on the ActionMessage class as a static property?

nigel-sampson:
I'd suggest making it a global fun with some relevant parameters, this way the behavior could be dynamic based on the inputs.

If you're making this on by default then I'd suggest adding some good docs around the change in behavior for the migration documentation.

vb2ae:
@KasperSK keeping the same behavior by default is probably the best way to do it

vb2ae:
@KasperSK should we Merge this in?

KasperSK:
I believe that I still need to make it so that you can change between the old and the new behavior. Might look at it this week. Then we should be able to merge this in.

PR #742, opened 2021-02-26
Build for NET5.0 is missing preprocessor symbol

BartoszPierzchlewiczMacrix:
.NET5.0 is .NET Core successor.
Without this preprocessor declaration binary for .NET5.0 has compiled code which is designed to work for .NET Framework.
That cause many unexpected errors.
Please take a look at https://docs.microsoft.com/en-us/dotnet/core/tutorials/libraries.
Build system is aware of  predefined Target Frameworks preprocessor declaration, and it will be better to use them in whole project instead creating own (which has different meaning).


nigel-sampson:
It looks the builds are currently failing due to an expired cert in the uwp projects.

BartoszPierzchlewiczMacrix:
Hi, @nigel-sampson what do you think about merging my PR.
Doesn't it looks good to You?

nigel-sampson (approved):

nigel-sampson:
Apologies, this slipped through the cracks.

PR #747, opened 2021-04-21
Update setup projects

vb2ae:
Updated setup projects to work with calibur.micro 4


vb2ae:
@KasperSK I created an PR with updated setup projects

KasperSK:
I am in the process of reviewing it, looks good so far.

KasperSK:
@vb2ae The Setup solution looks good, I had no problems there in the Features solution I could not get the android project to run it said that IoC was not initialized, looks like the project is missing the setup of the IoC in the caliburn application so that might have been broken all along.
 In the Scenario solution I could not run the functional sample but I doubt that one is important.

KasperSK (approved):
Setup project works

PR #750, opened 2021-05-10
Update setup projects

vb2ae:
Added VB setup projects.   Updated setup projects to use the released version of caliburn.micro


KasperSK (commented):
I don't think you have to reference the `Caliburn.Micro.Platform.Core `as it should be bundled in the `Caliburn.Micro.Platform` and `Caliburn.Micro.Platform.Xamarin.Forms` packages.

vb2ae:
@KasperSK updated the setup projects

PR #757, opened 2021-05-21
Moved FrameAdapter classes to separate files

neilt6:
I moved the `FrameAdapter` classes to their own files, rather than being combined with the `INavigationService` interfaces.


neilt6:
@nigel-sampson For your consideration.

vb2ae:
@neilt6 I will look at it tonight.  I do like the idea of not having more than one class in a file

vb2ae:
it looks good

vb2ae (approved):
change looks good

Issue #760, opened 2021-06-11
Unable to display root view by passing the type as a non-generic parameter


godrose:
First of all, huge thanks for releasing this version.
I noticed few breaking changes.

`DisplayRootViewFor` now requires generic parameter of type instead of a regular one.
Please support this ability too.

vb2ae:
Not exactly sure what you mean by a regular one.  Could you post a code sample of what you would like to work?

godrose:
`DisplayRootViewFor(Type rootObjectType)`
as opposed to
`DisplayRootViewFor<TRootObject>()`

KasperSK:
#761 Fixes this but renames the method to follow the async theme.

PR #761, opened 2021-06-12
Add DisplayRootViewForAsync<T> for net 461 and dot net core

vb2ae:
Adds DisplayRootViewForAsync<T> to .net framework and .net core wpf projects.  This is related to issue #760

### DIFF
--- a/samples/features/Features.NetCore/Bootstrapper.cs
+++ b/samples/features/Features.NetCore/Bootstrapper.cs
@@ -41,9 +41,9 @@ namespace Features.CrossPlatform
                .PerRequest<NavigationTargetViewModel>();
         }
 
-        protected override void OnStartup(object sender, StartupEventArgs e)
+        protected override async void OnStartup(object sender, StartupEventArgs e)
         {
-            DisplayRootViewFor<ShellViewModel>();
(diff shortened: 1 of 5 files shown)


KasperSK (commented):
@vb2ae looks good, only thing I thought about was should we remove the method `DisplayRootViewFor` and only keep the `DisplayRootViewForAsync`?

vb2ae:
@KasperSK I removed the DisplayRootFor from .net 4.61 and .net core.   Updated feature sample

godrose:
@vb2ae @KasperSK 
How does the PR fix the bug?
The method is still missing

KasperSK:
@godrose as a part of the move to async a lot of methods was changed to include the async in the name. `DisplayRootViewFor(Type rootObjectType)` was renamed `DisplayRootViewForAsync(Type rootObjectType)` and `DisplayRootViewFor<TRootObject>()` was not so the error here was that `DisplayRootViewFor<TRootObject>()` should have changed to `DisplayRootViewForAsync<TRootObject>()`.

So both the generic and the method that take a type is there they are just renamed with the 'Async' in the end of there names.

vb2ae:
@godrose a nuget package with the change has not been released yet.

PR #763, opened 2021-06-21
Added async eventhandler to IActivate

### LINKED ISSUE #762: IDeactivate has AsyncEventHandler but IActivate has EventHandler?
Hi & thanks for your work on this great little framework!

I just spent some frustrating time debugging a race condition in my app.
I have a parent Screen which has child Screens, and I use ScreenExtensions.ConductWith to ensure that the children are activated/deactivated along with the parent.

My expectation was that a call to "await parentScreen.ActivateAsync()" wouldn't complete until all children were activated too. However, that's not the case - I think because ScreenExtensions.ActivateWith() fires off ActivateAsync() on the children without awaiting the result.
By contrast, "await parentScreen.DeactivateAsync()" DOES ensure that all children are deactivated before returning. That's because the Deactivated event handler is an AsyncEventHandler, and the DeactivateWith() method awaits the child DeactivateAsync() calls.

Further, it seems that Conductor.Collection.AllActive and Conductor.Collection.OneActive behave according to my expectation - so it's just Screen.ConductWith() that has the race condition.

Is there a reason that the behaviour differs between Activate and Deactivate? As a workaround, I'm considering subclassing Screen into something like "ConductingScreen", and using that instead of ConductWith(). But I would prefer to it to be fixed in the framework if possible.

KasperSK:
This fixes #762, and makes it so that `ConductWith` has the same behavior for `Activate` and `Deactivate`.

### DIFF
--- a/src/Caliburn.Micro.Core.Tests/ScreenExtentionTests.cs
+++ b/src/Caliburn.Micro.Core.Tests/ScreenExtentionTests.cs
@@ -37,6 +37,10 @@ namespace Caliburn.Micro.Core.Tests
 
             await ScreenExtensions.TryActivateAsync(root).ConfigureAwait(false);
 
+            Assert.True(child1.WasActivated, "child 1 should be active");
+            Assert.True(child2.WasActivated, "child 2 should be active");
+            Assert.True(child3.WasActivated, "child 3 should be active");
+
             await ScreenExtensions.TryDeactivateAsync(root, true).ConfigureAwait(false);
 
             Assert.True(child1.IsClosed, "child 1 should be closed");
@@ -55,7 +59,8 @@ namespace Caliburn.Micro.Core.Tests
             };
             var child2 = new StateScreen(TimeSpan.FromSeconds(3))
             {
-                DisplayName = "screen2"
+                DisplayName = "screen2",
+                IsClosable = false,
             };
 
             var child3 = new StateScreen()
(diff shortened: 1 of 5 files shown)


nigel-sampson:
This will be a breaking change so how you manage this in terms of sem ver will be interesting.

vb2ae:
@nigel-sampson you are right this is a breaking change.  Do you think we should have both Activated and ActivatedAsync events? We could mark Activated as obsolete and remove it in a future release.

    event AsyncEventHandler<ActivationEventArgs> ActivatedAsync;

    [Obsolete]
    event EventHandler<ActivationEventArgs> Activated;

nigel-sampson:
It's a tough one, because that style approach means you're left with `ActivatedAsync` as the name which creates conflict with the other event names which would lead to another round of deprecation / rename.

Another possibility is to hold off on fixing this till the next major version.

KasperSK:
@nigel-sampson I thought about this being breaking, however I also see it as something I should have taken into account when making the change for `IDeactivate`. Bumping the major version just for this feature would not be right so I guess this will have to wait till we have enough changes to justify `5.0`.

vb2ae (approved):
The changes look good here.  @KasperSK  do you want to include this with V5?

KasperSK:
Yes @vb2ae, this needs to go in V5 since we where holding back on this change because it was breaking.

PR #765, opened 2021-06-24
Check for multiple assemblies being added to the cache

### LINKED ISSUE #737: Application crashes if scanned assemblies are duplicated
We have a plugin model that discovers view/viewmodels from external assemblies. If by mistake the assemblies are copied twice (to different discovery locations), the application would blow up with the following stack trace.

The `SelectedAssemblies()` method uses assemblies where as underlying cache which is the cause of the problem uses the Type name as the key.

```
System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Caliburn.Micro.EnumerableExtensions.Apply[T](IEnumerable`1 enumerable, Action`1 action)
   at Caliburn.Micro.AssemblySourceCache.<>c.<Install>b__3_0(Object s, NotifyCollectionChangedEventArgs e)
   at System.Collections.ObjectModel.ObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
   at Caliburn.Micro.BindableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
   at Caliburn.Micro.BindableCollection`1.OnUIThread(Action action)
   at Caliburn.Micro.BootstrapperBase.StartRuntime()
   at Caliburn.Micro.BootstrapperBase.Initialize()
   at ServiceInsight.Startup.AppBootstrapper..ctor() in /_/src/ServiceInsight/Startup/AppBootstrapper.cs:line 41
```

A simple fix would be to not add duplicate items to the cache if they already exist.

HEskandari:
Resolves #737

### DIFF
--- a/src/Caliburn.Micro.Platform.Core/AssemblySource.cs
+++ b/src/Caliburn.Micro.Platform.Core/AssemblySource.cs
@@ -90,7 +90,13 @@ namespace Caliburn.Micro
                     case NotifyCollectionChangedAction.Add:
                         e.NewItems.OfType<Assembly>()
                             .SelectMany(a => ExtractTypes(a))
-                            .Apply(t => TypeNameCache.Add(t.FullName, t));
+                            .Apply(t =>
+                            {
+                                if (!TypeNameCache.ContainsKey(t.FullName))
+                                {
+                                    TypeNameCache.Add(t.FullName, t);
+                                }
+                            });
                         break;
                     case NotifyCollectionChangedAction.Remove:
                     case NotifyCollectionChangedAction.Replace:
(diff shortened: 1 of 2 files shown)


vb2ae:
looks good

KasperSK:
Looks like the build Pipeline failed I will have a look at it.

KasperSK (commented):
Would it make sense to add this to the Reset case as well?

HEskandari:
@KasperSK Yeah, I'll add that, but TBH, I don't know how that bit of the code is supposed to work?

Calling `Clear` on `AssemblySource.Instance` calls back the `CollectionChanged` event with the Rest flag, but at that point, the `AssemblySource.Instance` is empty, so that code in the Reset block (except for the cache clear) that tries to call the data back into the cache is essentially a no-op.

KasperSK:
@HEskandari `BindableCollection` sends a `Reset` if you do an `AddRange`, `RemoveRange` or `Refresh` on it. It is due to some WPF behavior.

HEskandari:
@KasperSK yep, figured out :) Pushed an update.

PR #780, opened 2021-10-24
Net 6 0 and Visual Studio 2022 support

vb2ae:
Net 6 0 and Visual Studio 2022 support.

Updated UAP version so Caliburn Micro can be built in VS 2022 RC

Added support for Default implementation of methods to the IHandle interface for .net core 3.1, .net 5, and .net 6

Added .net 6 support


KasperSK:
Looks like its the Feature solution that breaks the build.

KasperSK:
I think the UAP version in the Features project need to be bumped to uap10.0.18362, and the reference path need to be updated @vb2ae should we get this done so that we can make a release that supports .net6?

vb2ae:
@KasperSK  fixed the build.  Yes we can do a .net 6 release.   Did you get a chance to look at documentation changes you requested?  Would be nice to release that too.

PR #815, opened 2022-07-16
Handle application lifetime events.

Stannieman:


src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/BootstrapperBase.cs
```diff
@@ -102,13 +101,12 @@ protected virtual void StartRuntime()
         /// </summary>
         protected virtual void PrepareApplication()
         {
-            //TODO: (Avalonia) Implement application events when Avalonia supports them
-            
-            //Application.Startup += OnStartup;
-
-            //Aplication.DispatcherUnhandledException += OnUnhandledException;
+            if (Application.Current.ApplicationLifetime is IControlledApplicationLifetime controlledApplicationLifetime)
```
Stannieman, line 104: This is potentially confusing.
For iOS, Android and Blazor apps this is ISingleViewApplicationLifetime and it does not have the events. So the user could override the methods and think they will be called but they never will.

What do we think about this?

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/BootstrapperBase.cs
```diff
@@ -102,13 +101,12 @@ protected virtual void StartRuntime()
         /// </summary>
         protected virtual void PrepareApplication()
         {
-            //TODO: (Avalonia) Implement application events when Avalonia supports them
-            
-            //Application.Startup += OnStartup;
-
-            //Aplication.DispatcherUnhandledException += OnUnhandledException;
```
Stannieman, line 109: Avalonia indeed does not seem to have something like this, nor is it planned. But I think adding a try/catch in Program.cs is close enough.

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/BootstrapperBase.cs
```diff
@@ -159,6 +157,24 @@ protected virtual void BuildUp(object instance)
...
+        protected virtual void OnStartup(object sender, ControlledApplicationLifetimeStartupEventArgs e)
+        {
+        }
+
+        /// <summary>
+        /// Override this to add custom behavior on exit.
+        /// </summary>
+        /// <param name="sender">The sender.</param>
+        /// <param name="e">The event args.</param>
+        protected virtual void OnExit(object sender, ControlledApplicationLifetimeExitEventArgs e)
```
Stannieman, line 174: In the other BootstrapperBase this is EventArgs instead of ExitEventArgs, but for OnStartUp it uses the more specific args. Was there a reason to use EventArgs for exit or is it ok to have ControlledApplicationLifetimeExitEventArgs like this?

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/BootstrapperBase.cs
```diff
@@ -102,13 +101,12 @@ protected virtual void StartRuntime()
         /// </summary>
         protected virtual void PrepareApplication()
         {
-            //TODO: (Avalonia) Implement application events when Avalonia supports them
-            
-            //Application.Startup += OnStartup;
-
-            //Aplication.DispatcherUnhandledException += OnUnhandledException;
+            if (Application.ApplicationLifetime is IControlledApplicationLifetime controlledApplicationLifetime)
```
Stannieman, line 104: This is potentially confusing.
For iOS, Android and Blazor apps this is ISingleViewApplicationLifetime and it does not have the events. So the user could override the methods and think they will be called but they never will.

What do we think about this?

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/BootstrapperBase.cs
vb2ae: ok

src/Caliburn.Micro.Avalonia/Platforms/netcore-avalonia/BootstrapperBase.cs
vb2ae: we can try it

PR #816, opened 2022-07-26
Expose ShowDialogAsync in IWindowManager.

Stannieman:


Stannieman:
InforOwnerOf has a check for PresentationSource.FromVisual(application.MainWindow) that is commented.
Any idea what it is for? To check that MainWindow was actually shown?

PR #826, opened 2022-08-26
Fix bindings that are not working with net6.0

### LINKED ISSUE #820: CanSayHello doesn't work view-to-VM communication from Caliburn startup documentation
I am following the instructions here using WPF dotnet 6.0 project.
https://caliburnmicro.com/documentation/configuration

Everything works but setting the CanSayHello bool for the SayHello Command. The issue is the CanSayHello is not triggered when the textbox changes. Caliburn convention should allow that to be triggered since it has the same name with the "Can" in front of it

Edit:
it fails only for the 4.0.210 version, it works for 3.2.0. Something is broken from the last release

thanks

KasperSK:
Fixes #820


vb2ae (approved):
looks good

PR #829, opened 2022-09-11
Winappsdk

vb2ae:
first draft.  still need to work on navigation service and remove a lot of debugging messages


KasperSK:
Is there a reason why the 'Semi' shared platform files are copied instead of linked into the WinAppSdk csproj? Should the name be  WinUI3?

vb2ae:
I plan to move them when it is working.  I thought it would be easier to debug this way

Issue #836, opened 2022-11-27
Remove .net 5 support 


vb2ae:
Remove support for .net 5 since it is no longer supported by microsoft

PR #837, opened 2022-11-27
remove .net 5. and add support for .net 7

vb2ae:
remove .net 5. and add support for .net 7 update global.json for version of visual studio 2022 version 17.4

Addresses issue #836 and issue #838


samples/features/Features.NetFive/Features.NetFive.csproj
```diff
@@ -2,7 +2,7 @@
```
KasperSK, line 2: Should we rename this project to Net instead of NetFive?

vb2ae: That is a good Idea I will work on that

vb2ae: updated

kfordmad (commented):
This seems to tick the normal checkboxes.  @KasperSK is there anything specific needed here.  I would happily work on this.

vb2ae:
@kfordmad we just need to get the pr approved and we can merge it into the master branch

kfordmad:
> 

Yeah.  It seems to give me the option to approve, but I suspect it wont actually go through, and I am not a member of the project, so I figured I would poke.  Was very interested in using this in a new project, and was hoping to use 7.  Suppose I can fork and self compile until this is approved.

KasperSK (approved):

Issue #831, opened 2022-10-05
Caliburn Micro for Maui


schovan:
I started experimenting with Maui and found that other MVVM frameworks for MAUI are either unusable (haven't features that I want) or bugged.
Can I somewhere download alpha / beta build for MAUI and test it?
Do you plan some pre-release / when ?

406841895:
![image](https://user-images.githubusercontent.com/38776389/195487286-eb04dc4a-e938-414c-8170-7fa9cebdee76.png)
The color of the title bar is not normal, and the background color of the navigation bar is wrong.
The program crashes when exiting debug for the first time

schovan:
@406841895 Is this a version you made it by yourself or you downloaded it somewhere?

vb2ae:
Not sure what you are asking?

schovan:
@vb2ae me neither :-) maybe caliburn micro 5 beta? Is this https://github.com/Caliburn-Micro/Caliburn.Micro/releases/tag/5.0.19-beta%2B7e85e48931 what I should use?

vb2ae:
yes the caliburn micro 5 beta has a maui version.  would love any feedback.  It is available in our myget feed

406841895:
> @406841895 Is this a version you made it by yourself or you downloaded it somewhere?

This app is written based on the Microsoft example, I made a mistake and now the style is fine, but I don't know how to configure the IOC container

406841895:
> yes the caliburn micro 5 beta has a maui version. would love any feedback. It is available in our myget feed

In the maui version, why is there no AppBootstrapper, and how do I configure the IOC container SimpleContainer

KasperSK:
Try looking at this sample for inspiration on how to setup Caliburn micro with MAUI:
[MAUI Sample](https://github.com/Caliburn-Micro/Caliburn.Micro/tree/master/samples/setup/Setup.Maui)

KasperSK:
I am aware of a bug the I need to get around to fixing it has to do with the viewlocator and namespace mappings. If you run into issues finding views it is because CM is trying to use Activity as a suffix instead of view.

schovan:
@KasperSK Has anyone tested if dependency injection really works for MAUI? Looks like it only works for parametless constructor. Look at this file for example: https://github.com/Caliburn-Micro/Caliburn.Micro/blob/master/src/Caliburn.Micro.Platform/Platforms/Maui/Windows/CaliburnApplication.cs
[...]

PR #839, opened 2022-12-12
Maui dependency injection

vb2ae:
Update the maui implementation to do dependency injection in Caliburn Micro MauiApplication instead of platform specific CaliburnApplication

Related to #831

### DIFF
--- a/samples/setup/Setup.Android/Resources/Resource.Designer.cs
+++ b/samples/setup/Setup.Android/Resources/Resource.Designer.cs
@@ -14,7 +14,7 @@ namespace Setup.Android
 {
 	
 	
-	[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Xamarin.Android.Build.Tasks", "12.3.99.16")]
+	[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Xamarin.Android.Build.Tasks", "13.1.0.5")]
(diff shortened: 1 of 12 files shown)


KasperSK:
What happens if I want to register platform specific services into the IoC container when it is structured like this?

vb2ae:
As long as you use the same instance of the simple container in the platforms it work properly

vb2ae:
Here is an example 

In maui app.xaml.cs 

       private static SimpleContainer _container;
        public static SimpleContainer container
        {
            get
            {
                if (_container == null)
                {
                    _container = new SimpleContainer();
                    _container.Instance(_container);
                }
                return _container;
            }
        }
        protected override void Configure()
        {


            container.PerRequest<IDataService, DataService>();
            container.PerRequest<MainViewModel>();
            container.PerRequest<MainView>();
        }


In windows app.xaml.cs 

        protected override void Configure()
        {
           ClientNoSqlDB.Samples.Maui.App.container.PerRequest<IPlatformSpecific, WindowsInstance>();
        }

In android.app.xaml

        protected override void Configure()
        {

            ClientNoSqlDB.Samples.Maui.App.container.PerRequest<IPlatformSpecific, AndroidSpecific>();
        }

image

image

KasperSK (approved):

PR #845, opened 2023-02-11
Update to Avalonia 11.0.0-preview5.

Stannieman:
For reference: https://github.com/AvaloniaUI/Avalonia/wiki/Breaking-Changes


vb2ae:
Thank you for doing this.  Think you need to update the Avalonia ui nuget packages reference in setup Avalonia ui app.

Stannieman:
Done! The Fluent theme is in its own package now and light/dark has to be set in a different way.

PR #847, opened 2023-03-16
Create default arguments depending on actual InitializeComponent para…

Stannieman:
…meters.

### DIFF
--- a/src/Caliburn.Micro.Platform/ViewLocator.cs
+++ b/src/Caliburn.Micro.Platform/ViewLocator.cs
@@ -509,16 +509,23 @@ namespace Caliburn.Micro
         {
 #if XFORMS
             return;
-#elif !WINDOWS_UWP
-            var method = element.GetType()
-                .GetMethod("InitializeComponent", BindingFlags.Public | BindingFlags.Instance);
-
-            method?.Invoke(element, null);
-
-#else
+#elif WINDOWS_UWP
             var method = element.GetType().GetTypeInfo()
                 .GetDeclaredMethods("InitializeComponent")
                 .SingleOrDefault(m => m.GetParameters().Length == 0);
+
+            method?.Invoke(element, null);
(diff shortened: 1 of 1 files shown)


src/Caliburn.Micro.Platform/ViewLocator.cs
```diff
@@ -509,16 +509,23 @@ public static void InitializeComponent(object element)
...
+
+            method?.Invoke(element, null);
+#elif AVALONIA
             var method = element.GetType()
                 .GetMethod("InitializeComponent", BindingFlags.Public | BindingFlags.Instance);
 
-            method?.Invoke(element, null);
+            var arguments = Enumerable.Repeat(Type.Missing, method.GetParameters().Length).ToArray();
 
+            method?.Invoke(element, arguments);
```
Stannieman, line 524: Avalonia started generating InitializeComponents using a source generator.
https://github.com/AvaloniaUI/Avalonia/blob/1da8891d846cbffd3c1f7ca6a837ea6e457a4bfc/src/tools/Avalonia.Generators/NameGenerator/InitializeComponentCodeGenerator.cs

If your project references `Avalonia.Diagnostics` then its signature will look like
`public void InitializeComponent(bool loadXaml = true)`
otherwise it will look like
`public void InitializeComponent(bool loadXaml = true, bool attachDevTools = true)`

Note that even if attachDevTools is true this will only actually a in ttach them in debug builds.
I think in both cases it is reasonable for Caliburn to just use the default arguments as Avalonia generated.

The original code here called it always with no arguments which lead to runtime exception.
My change always calls it using the correct number of arguments and uses the default specified in the sigmature. I suspect it is still not working if some parameters do not have defaults, but for now the 2 cases I listed are the only possibilities.

Stannieman:
Build fails because some AvaloniaApplication1 example project sneaked into the solution file that does not exist in the repo.

PR #846, opened 2023-03-16
WIP: Update Avalonia support

danwalmsley:
@vb2ae I am from the Avalonia UI team! This PR targets your exising branch:
1. merges in master latest changed.
2. I have removed the need for ReactiveUI as your sample shouldnt depend on another mvvm framework (maybe I miss understood the need for reactiveui, let me know if this is a mistake)

Im continuing to test so I may add other fixes, and I will also test with latest upstream changes from Avalonia.

PR #848, opened 2023-03-18
Feature/rebase on master

Stannieman:


src/Caliburn,Micro.Platform.Core.Tests/Caliburn,Micro.Platform.Core.Tests.csproj
```diff
@@ -4,8 +4,8 @@
     <TargetFramework>net6.0</TargetFramework>
     <RootNamespace>Caliburn_Micro.Platform.Core.Tests</RootNamespace>
     <Nullable>enable</Nullable>
-
     <IsPackable>false</IsPackable>
+    <LangVersion>10</LangVersion>
```
Stannieman, line 8: Added this so it does not fail on unsupported Nullable setting.

Stannieman:
Build fails on CI one of the dotnet test commands. It succeeds on my machine. Not sure what is wrong with that.

src/Caliburn,Micro.Platform.Core.Tests/Caliburn,Micro.Platform.Core.Tests.csproj
vb2ae: I think changing nullable to disable will fix the issue.

Mrxx99: It looks like the problem is that this test project (Caliburn,Micro.Platform.Core.Tests) targets net6.0 while there is a global.json requiring SDK 7.0.100. But what does this test project do anyway? It does contain no tests (has a spelling error in it's name) and does not exist on master. I guess it could simply be removed.

vb2ae: it might not hurt to update the microsoft.tests.sdk to 17.5.0 also

vb2ae: Looking at the test project that is failing.  I do not see any tests in it.  You could remove it until you add the tests you wanted

Stannieman: Ok I'll remove that project tonight. Also won't add tests to it anyway cause I did not create this project in the first place 🤷‍♂️

Regarding updating test SDK. Is this the right moment or PR for that? My main intention was just getting the branch up to date with master before looking into the other Avalonia issues still left. But if you want to have the new version I can bump it of course.

IMO the projects could use some love in general. I noticed LangVersion 9, 10 and latest used + some other settings that are not in sync. They could all be extracted to the root props file.

vb2ae: I can update the test sdk version in another pr.

Stannieman: We could just close this PR and focus on this one: https://github.com/Caliburn-Micro/Caliburn.Micro/pull/846
It has the same changes + some more relevant things.

PR #849, opened 2023-03-24
Do not call InitializeComponent for Avalonia.

Stannieman:
My fix from before should be removed. According to danwalmsley we don't need to call it, the default constructor of a control will do it for us. And if a user decides to use a non-default constructor then it is his responsibility to do it at the right moment. He added the `|| AVALONIA` on line 510 for that reason.


Stannieman:
Hmm crap maybe it is actually needed, checking with Avalonia

PR #852, opened 2023-04-05
Avalonia -> dont call InitialiseComponent, apps crash otherwise

danwalmsley:
Recent changes meant that initialisecomponent was getting called twice, once by the call generated by avalonia from the view ctor, and again via reflection, causing apps to crash.

### DIFF
--- a/src/Caliburn.Micro.Avalonia/Caliburn.Micro.Avalonia.csproj
+++ b/src/Caliburn.Micro.Avalonia/Caliburn.Micro.Avalonia.csproj
@@ -8,7 +8,7 @@
     <AssemblyOriginatorKeyFile>.\..\Caliburn.Micro.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <LangVersion>10</LangVersion>
-    <AvaloniaVersion>11.0.999-cibuild0032174-beta</AvaloniaVersion>
+    <AvaloniaVersion>11.0.999-cibuild0032488-beta</AvaloniaVersion>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -22,7 +22,6 @@
     <PackageReference Include="Avalonia.Markup.Xaml.Loader" Version="$(AvaloniaVersion)" />
     <PackageReference Include="Avalonia.Xaml.Behaviors" Version="11.0.0-preview6" />
     <PackageReference Include="Avalonia.Xaml.Interactivity" Version="11.0.0-preview6" />
-    <PackageReference Include="System.Reactive" Version="5.0.0" />
     <ProjectReference Include="..\Caliburn.Micro.Core\Caliburn.Micro.Core.csproj" />
     <ProjectReference Include="..\Caliburn.Micro.Platform.Core\Caliburn.Micro.Platform.Core.csproj" />
   </ItemGroup>
--- a/src/Caliburn.Micro.Platform/ViewLocator.cs
+++ b/src/Caliburn.Micro.Platform/ViewLocator.cs
@@ -510,7 +510,7 @@ namespace Caliburn.Micro
         /// <param name = "element">The element to initialize</param>
         public static void InitializeComponent(object element)
         {
-#if XFORMS
+#if XFORMS || AVALONIA
             return;
 #elif WINDOWS_UWP
             var method = element.GetType().GetTypeInfo()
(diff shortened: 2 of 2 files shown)


danwalmsley:
@Stannieman I know you changed this, however in the latest avalonia, the initialisecomponent call is automatically generated in the view ctor, so it will be called implicitly when the instance is created.

In my test apps, it was crashing because xaml was being processed twice.

Let me know though if you think this should be called in some cases that im unaware of though.

Stannieman:
@danwalmsley will check against latest Avalonia nightly.

Stannieman:
It is not working for me.
I changed my DI container to return "new SomeView()" which gets called when Caliburn resolves the view. SomeView is derived from UserControl. The only way to get something to show is creating a constructor and call InitializeComponent from there. My views have no other constructors.

But I wonder, these calls to InitializeComponent are not generated at build time, are they? Maybe something goes wrong in that step.

Stannieman:
There might be a misunderstanding. The VS template indeed generates a constructor that calls InitializeComponent, but when you remove that it is not called automagically anymore.

E.g. this simple app does not show the red background of MyView.
```
<Window xmlns="https://github.com/avaloniaui"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
        x:Class="AvaloniaCaliburnMinimalRepro.MainWindow"
        Title="AvaloniaCaliburnMinimalRepro"
        xmlns:root="clr-namespace:AvaloniaCaliburnMinimalRepro">
    <root:MyView></root:MyView>
</Window>
```

```
using Avalonia.Controls;

namespace AvaloniaCaliburnMinimalRepro
{
    public partial class MainWindow : Window
    {
        public MainWindow()
        {
            InitializeComponent();
        }
    }
}
```

```
<UserControl xmlns="https://github.com/avaloniaui"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
             mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
             x:Class="AvaloniaCaliburnMinimalRepro.MyView"
             Background="Red">
</UserControl>
```

```
using Avalonia.Controls;

namespace AvaloniaCaliburnMinimalRepro
{
    public partial class MyView : UserControl
    {
        // Uncomment this and it shows, comment it and it does not.
        //public MyView()
        //{
        //    InitializeComponent();
        //}
    }
}
```

Is there something on the control that we could check before calling InitializeComponent to know if it already happened or not?

Stannieman:
Also I am not sure why it crashes for you. What exception do you get?
I just tried calling InitializeComponent multiple times in the constructor and it does not crash for me.

danwalmsley:
@Stannieman what is the reason to remove the ctor? Avalonia expects the ctor to be left intact?

Stannieman:
The same is true for WPF. It generates this but you can remove it in which case Caliburn still handles it for you. Not sure if we want to change this Caliburn behavior.

Although in WPF it is convenient because you can just leave away the entire code behind file, for Avalonia that does not work.

KasperSK:
Does `InitializeComponent()` not protect itself from being called multiple times in `Avalonia`?

Stannieman:
For me it seems to protect (or at least not explode), I can call it as much as I want. No crashes and tried with and without Avalonia.Diagnostics dependency added (it the parameters of InitializeComponent).
I'd like to know what exception @danwalmsley is seeing.

KasperSK:
That would be my expectation, that it should protect itself, as for the `ctor` with more parameters add based on which dependencies are included that I do not like. Why not have it be an additional constructer instead of replacing the one parameter `ctor`?

Stannieman:
@KasperSK It comes from here: https://github.com/AvaloniaUI/Avalonia/blob/1da8891d846cbffd3c1f7ca6a837ea6e457a4bfc/src/tools/Avalonia.Generators/NameGenerator/InitializeComponentCodeGenerator.cs

TBH I do not have too much problems with this particular case.

KasperSK:
@Stannieman It might be okay for this, but looking at the code in your link it does not look like the method guards getting called twice? Perhaps there is logic somewhere else to handle the guarding?

Stannieman:
I assume AvaloniaXamlLoader.Load guards, but that actually calls into compiled XAML so it is harder to find out how that works.

vb2ae:
@danwalmsley any update on when new Avalonia nuget package will be available?

danwalmsley:
> is Caliburn beh

It wont work with dependency injection!

danwalmsley:
> Does Avalonia have any navigation features build in like the `Frame` class in WPF and UWP?

<ContentControl Content="{Binding MyPage}" />

vb2ae:
@danwalmsley all dependency injection containers?  or just Simple Container?

danwalmsley:
@Stannieman 

This is why you cant call it by reflection! Stuff will break! :D

image


I discussed it with the rest of the team, and the conclusion is this:

Currently avalonia does not support `AvaloniaXamlLoader.Load(this)` (which is what initialisecomponet does)
more than once. Its not detected or protected against.

The scenario of removing the codebehind file from xaml or from removing the ctor from code behind is not supported.

KasperSK:
@danwalmsley is there a particular reason why there is no guard against calling `AvaloniaXamlLoader.Load(this)` twice?

PR #854, opened 2023-04-07
Switch ItemsControl.Items property to ItemsSource for compatibility with latest Avalonia nightly

Stannieman:


vb2ae (approved):
looks good

PR #856, opened 2023-04-25
Update Avalonia to Preview 7

danwalmsley:
simply updates to preview 7 release,

and fix critical issue where xaml parsing will explode.


danwalmsley:
Removed the reflection p/invoking of InitializeComponent, Avalonia works more like Xamarin in this case.

Iv informed in the other PR that Avalonia doesnt support this, as shown below,.

image

vb2ae (commented):
looks good

PR #859, opened 2023-05-25
Allow view to be notified

erik-hooper:
This is in relation to
https://github.com/Caliburn-Micro/Caliburn.Micro/issues/593. I agree with the issue but is a pretty big breaking change for us. This should allow the correct sequence of activation, activated callbacks.


vb2ae (approved):
Looks good

vb2ae (approved):
Looks good

Issue #789, opened 2021-12-24
Please use OnActiveAsync carefully


a44281071:
We need to implement the following functions:
- active a screen.
- open/switch a new view.
- refresh some dates by active/init.

We can't use OnActiveAsync to load data. This will Put off the view display before async method finish. The solution is OnViewLoaded, BUT it can't await.

vb2ae:
@a44281071 are you asking for OnViewLoaded to be an async event?

a44281071:
@vb2ae 
Maybe something like "Before init/activ" and "After init/active" is comfortable. 
Any better idea?

jkattestaart:
I always stayed away from de the activate stuff because it was not async in
previous versions and created a StartAsync Method in my BaseViewModel.
Only drawback is that i have to implicit call the StartAsync

Reagrds

John Kattestaart
***@***.***


Op ma 27 dec. 2021 om 04:58 schreef SamTi（三台） ***@***.***>:

a44281071:
Perhaps we need a series of tutorials to demonstrate the best way to use the new asynchronous methods (active, inactive, handle event)

Yinimi:
One problem I've found with OnActivateAsync is, it is called just before IsActive is set to true (see [Screen.ActivateAsync()](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/a2854ab1f49f486e254255beec330818be79a204/src/Caliburn.Micro.Core/Screen.cs#L110-L111)).
This means if you have a main viewmodel that uses OnActivateAsync to initialize and activate some sub viewmodel like a login viewmodel then the OnActivateAsync in the sub viewmodel will be skipped [ConductorWithCollectionOneActive.ActivateItemAsync()](https://github.com/Caliburn-Micro/Caliburn.Micro/blob/a2854ab1f49f486e254255beec330818be79a204/src/Caliburn.Micro.Core/ConductorWithCollectionOneActive.cs#L70-L74). In this scenario it will never be called.

a44281071:
Please add some override methods for Active/Deactive. Like BeforActiveAsync/AfterActiveAsync..

theindra:
+1 to that. 

How can I activate a sub item from within `OnActivateAsync` of the parent?
I just upgraded from Caliburn 3.2 to 4 and this breaks.

nietras:
[...]

Issue #764, opened 2021-06-24
ConductorBaseWithActiveItem: View displayed before viewmodel active


madorjan:
Caliburn.Micro 4.0.173
.NET 4.8, WPF application

Is it possible to prevent a view (ActiveItem) from being displayed before its viewmodel is active (IsActive, OnActivateAsync finished)? In my opinion, its questionable if a view bound to an "unfinished" data context (viewmodel) makes sense anyway.

https://github.com/Caliburn-Micro/Caliburn.Micro/commit/ee6fdb1194883ebc61c8224f488b5449c97f0822#diff-0e35b8f45cdc997e2bc2226095d8a3e93342e65eca1162a4e412b1d6a6134448

introduced the current behavior.

However, its currently not possible to await some preprocessing or user-input (e.g. question-dialog before view opens) in OnActivateAsync or OnInitializeAsync without having an unfinished view displayed, or am I missing something?

Another related aspect would be, how to handle early exits in OnActivateAsync or OnInitializeAsync? Do I have to call TryCloseAsync if something fails or gets cancelled within these methods? Would this even work, since viewmodels would not be active at this point? How to get rid of the displayed view in such cases?

Thanks for your support!

lploumen:
Hi,
I got same problem while trying to update from 3.2.0 to 4.x.x.
I have a ViewModel (Conductor<IScreen>.Collection.OneActive) bound to a TabControl. In the TabControl VM OnActivateAsync method, I initialize the Items but their OnActivateAsync method never get called.

vb2ae:
@lploumen could you share some code which shows the issue?

PR #898, opened 2024-04-02
Fix activation for nested conductor / screen scenarios

Yinimi:
There have been some discussions and changes related to when OnActivate() is called.
#593 He wanted to use IsInitialized / IsActive to control the visibility of some loading indicator. If the property is set to true before the OnActivate / OnInitialize call it will obviously not be able to use it for determining if the loading has finished.
I can understand the reason for this request but I think a method called OnActivate() should be called after the IsActive is set to true, otherwise Activating or BeforeActivate are better names in my opinion.

Since some devs (like me) want to use OnActivate() to trigger the initializing of the ViewModel there are scenarios where it is problematic if the IsActive is not set yet. I created the [ActivateWhileActivateOneLevel()](https://github.com/Yinimi/Caliburn.Micro/blob/e246f672cd93b0340f3792eefdfa58d955c6a18e/src/Caliburn.Micro.Core.Tests/ConductorWithCollectionOneActiveTests.cs#L125) test method to demonstrate it. 
To provide such a method @erik-hooper added with #859 a [OnActivate**d**()](https://github.com/Caliburn-Micro/Caliburn.Micro/pull/859/commits/5a524b0904ac59dcb43425e286764a4e42619a7c) method which is called after the property has been set.
If I use this method with another test [ActivatedWhileActivatedOneLevel()](https://github.com/Yinimi/Caliburn.Micro/blob/e246f672cd93b0340f3792eefdfa58d955c6a18e/src/Caliburn.Micro.Core.Tests/ConductorWithCollectionOneActiveTests.cs#L148C27-L148C58) The issue is resolved.

I have a more complicated and nested scenario. [example Test](https://github.com/Yinimi/Caliburn.Micro/blob/e246f672cd93b0340f3792eefdfa58d955c6a18e/src/Caliburn.Micro.Core.Tests/ConductorWithCollectionOneActiveTests.cs#L171)
* Outer Conductor
  * some Page
  * Inner Conductor
    * View 1
    * View 2

In my app the inner conductor handles the navigation depending on some configuration.
The outer conductor handles basic stuff and also the display of the user preferences ("some Page").
If the View 1 now wants automatically to continue after the user preferences have been shown (think of auto login was just enabled) then in this nested scenario the View 2 will not be initialized because all this is triggered by the OnActivate() method of the inner conductor.

With this change the behavior should be like before the changes of #593 where the active item(s) are activated after the conductor has IsActive enabled.

In #789 it was also discussed to add something like BeforeActivate and AfterActivate.
I added a [OnInitializedAsync()](https://github.com/Yinimi/Caliburn.Micro/blob/b0cbfc64eca3da3e94d32db6149e816bf05089c1/src/Caliburn.Micro.Core/Screen.cs#L107) to adress this as well. I didn't touch the OnViewLoaded (it was wished to make this async), this looks more complicated.


vb2ae:
Yes the release is version 5

src/Caliburn.Micro.Core/ConductorWithCollectionOneActive.cs
vb2ae, line 1: I like the OnActivated name better but could you leave OnActivate and mark it as obsolete? This will keep the amount of code we break when someone updates the caliburn micro version in their project to a minimum

Yinimi:
I added the obsolete messages

Yinimi:
I think #764 Is also resolved with this change

a44281071:
asp.net Blazor use ```void OnAfterRender()``` and ```Task OnAfterRenderAsync()``` for view update. And use ```void OnInitialized()``` and ```Task OnInitializedAsync()``` for init load.

ASP.NET Core Razor component lifecycle:
> [Component initialization](https://learn.microsoft.com/en-us/aspnet/core/blazor/components/lifecycle?view=aspnetcore-8.0#component-initialization-oninitializedasync)

Yinimi:
If I read the build pipeline log correct the android 28 sdk is missing? Is the pipeline running on a specific machine with a build agent? Then it would probably be required to install it there manually

vb2ae:
Working on a fix for android version in another branch

src/Caliburn.Micro.Core.Tests/ConductorWithCollectionOneActiveTests.cs
```diff
@@ -24,6 +24,46 @@ protected override async Task OnDeactivateAsync(bool close, CancellationToken ca
...
+            {
+                await ChangeActiveItemAsync(newItem, closePrevious);
+            }
+        }
+
+        private class ActivateScreen : Screen
+        {
+            public event EventHandler RequestContinue;
+
+            public int ActivateCalledCount { get; private set; }
```
vb2ae, line 39: Is this just to see show the user how many times Activate is called?

src/Caliburn.Micro.Core/Screen.cs
```diff
@@ -173,14 +174,24 @@ public virtual async Task TryCloseAsync(bool? dialogResult = null)
         /// <summary>
         /// Called when initializing.
         /// </summary>
+        [Obsolete("Override OnInitializedAsync")]
```
vb2ae, line 177: Thanks for marking old method and marking obsolete

vb2ae (changes requested):
Looks good.  Could you please update the Feature apps with the change

Yinimi:
@vb2ae I updated the Feature branch.
Additionally I fixed the scenarios solution but it looked very outdated, it was still using some caliburn micro version 3 dlls.

vb2ae:
@Yinimi thank you the samples do need help.

Yinimi:
There seems to be a related issue #906 
The changes cause unit tests to fail

vb2ae:
@Yinimi I will look at it

src/Caliburn.Micro.Core/ConductorBaseWithActiveItem.cs
```diff
@@ -39,12 +39,13 @@ object IHaveActiveItem.ActiveItem
         /// <returns>A task that represents the asynchronous operation.</returns>
         protected virtual async Task ChangeActiveItemAsync(T newItem, bool closePrevious, CancellationToken cancellationToken)
         {
+            var previousItem = _activeItem;
```
vb2ae, line 42: Good find

vb2ae (approved):
Thank you for finding this

a44281071:
Maybe a 'state' like connection is better. Just whether one is activated or not expresses too little meaning.
Pending. InProgress. Completed, Failed...

For example, "In Progress" indicates that it is activating. 

For CM.v5, the func name 'activating' and 'Activated/AfterActivated' is better. We don't need 'Active' means it is being activated.
“Don't let the doubt persist”. Just tell us clearly in the tutorial documentation and let us replace it globally.

vb2ae:
@a44281071 could you create an issue for the requested change?

Issue #906, opened 2024-10-26
Async calls running on separate context in OnDeactivateAsync cause UI not to refresh view


darxis:
Caliburn.Micro NuGet package version 4.0.212.

I have found a bug in the screen Activation/Deactivation logic.

The problem is when a long-running task is run within `OnDeactivateAsync` that is not run on the captured context.

Please see the reproduction example I pushed to my fork - [https://github.com/darxis/Caliburn.Micro/tree/bug-async-calls-in-ondeactivateasync](https://github.com/darxis/Caliburn.Micro/tree/bug-async-calls-in-ondeactivateasync)

Please see commit #[f136656](https://github.com/darxis/Caliburn.Micro/commit/f136656ed173960f424ad3d306b31952024acba9)

I modified the `Caliburn.Micro.Tutorial.Wpf` sample code - I modified the `ShellViewModel` to display a `TabControl` for two ViewModels - one of them is simulating a long-running task not on the captured context - it's just an `await Task.Delay(...)`:
```c#
protected override async Task OnDeactivateAsync(bool close, CancellationToken cancellationToken)
{
  // Here we simulate some long-running task that runs on another context (it also could be for example running any long-running task using Task.Run)
  // This causes the bug to trigger - the view for 'SomeOtherViewMode' doesn't show when selecting the 'Some other view' tab
  // If the following line is removed everything works fine.
  await Task.Delay(1000, cancellationToken);
  // Or any of the following:
  // await Task.Run(async () => await SomeLongRunningTaskAsync(cancellationToken), cancellationToken);
  // await Task.Run(() => Thread.Sleep(1000), cancellationToken);

  await base.OnDeactivateAsync(close, cancellationToken);
}
```

Reproduction steps:
1. Run the application
2. Click on the 'Some other view' tab
3. The tab view is not displayed, but it should be

I think it what happens is the following:
1. When you click the 'Some other view' tab the Dispatcher (or some other thread) sets the `ActiveItem` with the `SomeOtherViewModel` instance. 
2. The `ActiveItem` setter call returns before the `OnDeactivateAsync` method is fully executed as the `ActivateItemAsync` call is not awaited so the `_activeItem` is not changed because it is updated after `OnDeactivateAsync` is completed. That means that the `ActiveItem` setter call returns but the `ActiveItem` would return the old viewmodel until the `OnDeactivateAsync` completes and `_activeItem` is changed.
[...]

PR #891, opened 2023-11-09
Create dotnet.yml

vb2ae:
add dotnet build want to start creating packages on github instead of myget

PR #907, opened 2024-10-26
Fix issue #906

darxis:
Fixes issue #906. Please review.

### DIFF
--- a/src/Caliburn.Micro.Core/ConductorBaseWithActiveItem.cs
+++ b/src/Caliburn.Micro.Core/ConductorBaseWithActiveItem.cs
@@ -39,11 +39,11 @@ namespace Caliburn.Micro
         /// <returns>A task that represents the asynchronous operation.</returns>
         protected virtual async Task ChangeActiveItemAsync(T newItem, bool closePrevious, CancellationToken cancellationToken)
         {
-            await ScreenExtensions.TryDeactivateAsync(_activeItem, closePrevious, cancellationToken);
-            newItem = EnsureItem(newItem);
-
             _activeItem = newItem;
             NotifyOfPropertyChange(nameof(ActiveItem));
+
+            await ScreenExtensions.TryDeactivateAsync(_activeItem, closePrevious, cancellationToken);
+            newItem = EnsureItem(newItem);
 
             if (IsActive)
                 await ScreenExtensions.TryActivateAsync(newItem, cancellationToken);
--- a/src/Caliburn.Micro.Core.Tests/ConductorWithCollectionOneActiveTests.cs
+++ b/src/Caliburn.Micro.Core.Tests/ConductorWithCollectionOneActiveTests.cs
@@ -24,6 +24,45 @@ namespace Caliburn.Micro.Core.Tests
             }
         }
 
+        private class AsyncActivationScreen : Screen
+        {
+            private readonly bool _simulateAsyncOnActivate;
+
+            private readonly bool _simulateAsyncOnDeactivate;
+
+            private readonly TimeSpan _simulateAsyncTaskDuration;
+
+            public AsyncActivationScreen(bool simulateAsyncOnActivate, bool simulateAsyncOnDeactivate,
+                TimeSpan simulateAsyncTaskDuration)
+            {
+                _simulateAsyncOnActivate = simulateAsyncOnActivate;
+                _simulateAsyncOnDeactivate = simulateAsyncOnDeactivate;
+                _simulateAsyncTaskDuration = simulateAsyncTaskDuration;
+            }
+
+            protected override async Task OnActivateAsync(CancellationToken cancellationToken)
+            {
+                await base.OnActivateAsync(cancellationToken);
+
+                if (_simulateAsyncOnActivate)
+                {
+                    // Task.Delay doesn't run within captured context
+                    await Task.Delay(_simulateAsyncTaskDuration, cancellationToken);
+                }
+            }
+
+            protected override async Task OnDeactivateAsync(bool close, CancellationToken cancellationToken)
+            {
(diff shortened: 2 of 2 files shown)


vb2ae (changes requested):
Charges look good.  Will need to merge in pr #891 once I get a review and merged into master.  It will fix the build issues.

src/Caliburn.Micro.Core/ConductorBaseWithActiveItem.cs
```diff
@@ -39,12 +39,12 @@ object IHaveActiveItem.ActiveItem
...
         protected virtual async Task ChangeActiveItemAsync(T newItem, bool closePrevious, CancellationToken cancellationToken)
         {
-            await ScreenExtensions.TryDeactivateAsync(_activeItem, closePrevious, cancellationToken);
-            newItem = EnsureItem(newItem);
-
             _activeItem = newItem;
             NotifyOfPropertyChange(nameof(ActiveItem));
 
+            await ScreenExtensions.TryDeactivateAsync(_activeItem, closePrevious, cancellationToken);
+            newItem = EnsureItem(newItem);
```
vb2ae, line 46: think you need to move this above _activeItem = newItem;

darxis: You're right, I overlooked this call. Commited [10180d1](https://github.com/Caliburn-Micro/Caliburn.Micro/pull/907/commits/10180d1d3b1d8a71716ecd919debe60b4eb957ea)

vb2ae: Are you still having issues with INotifyPropertyChanged?

darxis: Do you mean the `NotifyOfPropertyChange(nameof(ActiveItem));` call?

I meant that this call should have prevented the bug on the unfixed master branch because it should trigger a UI refresh, but it have not and I don't know why. Please check my [bug-async-calls-in-ondeactivateasync](https://github.com/darxis/Caliburn.Micro/tree/bug-async-calls-in-ondeactivateasync) branch where I modified the `Caliburn.Micro.Tutorial.Wpf` example to reproduce this bug.

darxis: > I think it what happens is the following:
> 1. When you click the 'Some other view' tab the Dispatcher (or some other thread) sets the `ActiveItem` with the `SomeOtherViewModel` instance. 
> 2. The `ActiveItem` setter call returns before the `OnDeactivateAsync` method is fully executed as the `ActivateItemAsync` call is not awaited so the `_activeItem` is not changed because it is updated after `OnDeactivateAsync` is completed. That means that the `ActiveItem` setter call returns but the `ActiveItem` would return the old viewmodel until the `OnDeactivateAsync` completes and `_activeItem` is changed.
> 3. The thread/code that set the `ActiveItem` marks that property as "Dirty" and triggers refreshing of that property before the underneath `_activeItem` field is changed - you can check that by putting a breakpoint on the `ActiveItem` setter.

But then the `OnDeactivateItem` finally completes asynchronously and the `_activeItem` is set and the `NotifyOfPropertyChange(nameof(ActiveItem));` should work, but it is not working and the UI still displays the old view.

vb2ae: Ok I will look

darxis: It seems like whatever the thing that should handle the `PropertyChanged(ActiveItem)` event and refresh the UI is just ignoring this event.

darxis: > Okay, I managed to find out why the `TabControl` doesn't update the UI despite `PropertyChanged(ActiveItem)` being called. It is because when you click the second tab on the UI to show it then it is the `TabControl` that sets the `ActiveItem` and internally `TabControl` will remember this newly selected tab as the active tab. And later when the `PropertyChanged(ActiveItem)` is called the `TabControl` compares if the active tab it remembered is different than the new one because if that's the case there is no need to do anything. In this case it is the same tab that's why the UI doesn't update to the newly selected tab view.
> 
> So `TabControl` has the correct tab remembered but it displays the view for the old tab as `ActiveItem/_activeItem` is not updated yet because it is updated asynchronously later.
> 
> Also this bug doesn't occur if you activate the other tab programatically by calling `ActiveItem = _someOtherViewModel` from within the `Conductor`, because TabControl doesn't know the new tab yet, it get to know it only when the `NotifyOfPropertyChange(nameof(ActiveItem))` is called and then it refreshes the UI with the correct view.
> 
> TL;DR - everything is fine with the `NotifyOfPropertyChange(nameof(ActiveItem));` call and `ActiveItem` should set its backed field `_activeItem` synchronously before exiting the setter, not asynchronously within a new `Task` - the PR I submitted fixes this behavior.

vb2ae:
pr #891 is merged in can you merge or rebase this to the latest in the master branch?  I will approve and merge in

vb2ae (approved):

vb2ae:
@darxis thanks It is merged in. I need to make a few changes to the build github action. Did not think about setting up the github feed credentials in the build from a fork would fail because the secrets are not available

darxis:
@vb2ae Thanks. Can we make a new v4 release with this bugfix? I am still using v4 and it would be nice to have it fixed in as v4 (v4.0.213?) as v5 is still in beta.

vb2ae:
There is 4.0.212 branch we can merge your fix into.   Not sure it will build because it uses some unsuported platforms
 https://github.com/Caliburn-Micro/Caliburn.Micro/tree/V4_0_212

PR #924, opened 2024-11-05
Make regex in Parser case-insensitive with a timeout

vb2ae:
Updated LongFormatRegularExpression in the Parser class to be case-insensitive by adding RegexOptions.IgnoreCase. Also set a timeout of 10 seconds using TimeSpan.FromSeconds(10) to prevent potential performance issues from long-running regex operations.

### DIFF
--- a/src/Caliburn.Micro.Platform/Parser.cs
+++ b/src/Caliburn.Micro.Platform/Parser.cs
@@ -47,7 +47,7 @@ namespace Caliburn.Micro
     /// </summary>
     public static class Parser
     {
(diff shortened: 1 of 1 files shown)


src/Caliburn.Micro.Platform/Parser.cs
KasperSK, line 1: Would this not change how Event and Action is parsed? So we match both Event and event? 

Would that not break later string replace that looks for Event in uppercase?

vb2ae:
good point going to close this pr

PR #934, opened 2024-12-11
Update version.json

vb2ae:
Update PublicReleaseSpec so GitVersion does not think it is prerelease and add release commit id to end of package version.

@darxis can you please review this

### DIFF
--- a/version.json
+++ b/version.json
@@ -5,7 +5,7 @@
         "semVer": 2
     },
     "publicReleaseRefSpec": [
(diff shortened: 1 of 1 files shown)


darxis (approved):
I've no experience with GitVersion but it looks ok. I can see that the PR build artifact packages have been labeled as `4.0.230-g6180972611` but I think it is because it is a PR build and not the `V4_0_212` branch build. Anyway I approve this.

PR #963, opened 2025-01-14
OnUIThread-Refactor and a few improvements

K4PS3:
This pull request to improve parts of the codebase for better readability and to fix the compiler warnings.
I think this pull request should be easy to review since the changes in the files are at minimum.
The code in "Platforms" can be improved, but will need extra work to be done.

You can take notes from this pull request without merging it, I don't mind that.

### **Summary of changes:**

- Added 'SafeRun' function in (BindableCollection.cs and PropertyChangedBase.cs) to remove [code duplication](https://code-maze.com/refactoring-dispensables-and-code-smells-in-csharp/).
- Added [documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags) in (RegExHelper.cs) for NameRegEx (Hover over NameRegEx to see the result).
- Changed action variables to [Local functions](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions).
- Changed [nested private classes](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/nested-types) to be **sealed**.
- Inlined the out parameters.
- Used [Pattern matching](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching) whenever applicable.
- Used [Compound assignment](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/assignment-operator#compound-assignment) whenever applicable.
- Used [nameof](https://grantwinney.com/using-nameof-to-avoid-magic-strings/) to avoid magic strings.
- Removed [this Keyword](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0003-ide0009) whenever possible.
- Removed [Zombie code](https://vfunction.com/blog/what-architects-should-know-about-zombie-code/).
- Removed [unnecessary using directives](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0005?pivots=lang-csharp-vb).
- Removed [#region preprocessor directives](https://enterprisecraftsmanship.com/posts/c-regions-is-design-smell/).
- Reformat Curly Braces
- Fixed couple of typos.

### SafeRun in Action
SafeRun

### Regex Documentation
regex

### DIFF
--- a/src/Caliburn.Micro.Core/BindableCollection.cs
+++ b/src/Caliburn.Micro.Core/BindableCollection.cs
@@ -43,14 +43,7 @@ namespace Caliburn.Micro
         {
             if (IsNotifying)
             {
-                if (PlatformProvider.Current.PropertyChangeNotificationsOnUIThread)
-                {
-                    OnUIThread(() => OnPropertyChanged(new PropertyChangedEventArgs(propertyName)));
-                }
-                else
-                {
-                    OnPropertyChanged(new PropertyChangedEventArgs(propertyName));
-                }
+                SafeRun(() => OnPropertyChanged(new PropertyChangedEventArgs(propertyName)));
             }
         }
 
@@ -59,21 +52,12 @@ namespace Caliburn.Micro
         /// </summary>
         public void Refresh()
         {
-            if (PlatformProvider.Current.PropertyChangeNotificationsOnUIThread)
-            {
-                OnUIThread(() =>
-                {
-                    OnPropertyChanged(new PropertyChangedEventArgs("Count"));
-                    OnPropertyChanged(new PropertyChangedEventArgs("Item[]"));
-                    OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
-                });
-            }
-            else
+            SafeRun(() =>
             {
                 OnPropertyChanged(new PropertyChangedEventArgs("Count"));
                 OnPropertyChanged(new PropertyChangedEventArgs("Item[]"));
                 OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
-            }
+            });
         }
 
         /// <summary>
@@ -83,14 +67,7 @@ namespace Caliburn.Micro
         /// <param name = "item">The item to be inserted.</param>
         protected override sealed void InsertItem(int index, T item)
         {
-            if (PlatformProvider.Current.PropertyChangeNotificationsOnUIThread)
-            {
-                OnUIThread(() => InsertItemBase(index, item));
-            }
-            else
-            {
-                InsertItemBase(index, item);
-            }
+            SafeRun(() => InsertItemBase(index, item));
         }
 
         /// <summary>
@@ -113,14 +90,7 @@ namespace Caliburn.Micro
         /// <param name = "item">The item to set.</param>
         protected override sealed void SetItem(int index, T item)
         {
-            if (PlatformProvider.Current.PropertyChangeNotificationsOnUIThread)
-            {
-                OnUIThread(() => SetItemBase(index, item));
-            }
-            else
-            {
-                SetItemBase(index, item);
-            }
+            SafeRun(() => SetItemBase(index, item));
         }
 
         /// <summary>
@@ -142,14 +112,7 @@ namespace Caliburn.Micro
         /// <param name = "index">The position used to identify the item to remove.</param>
         protected override sealed void RemoveItem(int index)
         {
-            if (PlatformProvider.Current.PropertyChangeNotificationsOnUIThread)
-            {
-                OnUIThread(() => RemoveItemBase(index));
-            }
-            else
-            {
-                RemoveItemBase(index);
-            }
+            SafeRun(() => RemoveItemBase(index));
         }
 
         /// <summary>
@@ -223,6 +186,7 @@ namespace Caliburn.Micro
                     InsertItemBase(index, item);
                     index++;
                 }
+
                 IsNotifying = previousNotificationSetting;
 
                 OnPropertyChanged(new PropertyChangedEventArgs("Count"));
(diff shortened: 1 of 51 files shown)


vb2ae (commented):
Thank you for this.  Thank you for fixing the typos in the comments.

src/Caliburn.Micro.Core/BindableCollection.cs
```diff
@@ -258,28 +215,34 @@ void RemoveRange()
...
+        }
+
+        /// <summary>
+        /// Executes the given action on the UI thread
+        /// </summary>
+        /// <remarks>An extension point for subclasses to customize how property change notifications are handled.</remarks>
+        /// <param name="action"></param>
+        protected virtual void OnUIThread(Action action) => action.OnUIThread();
+
+        private void SafeRun(Action action)
```
vb2ae, line 236: Rather than having 2 SafeRun functions why not create a new class in Caliburn.Micro.Core that has the function both BindableCollection and PropertyChangedBase can use.

K4PS3: Actually you don't have to create new class, you could just move the SafeRun to Execute static class and make it extension method.
I created the Two function in same file to make the refactor clear as steep 1 and make it is easy to review, (Refactor in same file).
Step 2 should be refactor the code duplication across files.

```
/// <summary>
///   Executes the action on the UI thread.
/// </summary>
/// <param name = "action">The action to execute.</param>
public static void OnUIThread(this Action action)
{
    PlatformProvider.Current.OnUIThread(action);
}

/// <summary>
/// Executes the action on safe manner either on UI thread or current Thread
/// </summary>
/// <param name="action"></param>
public static void SafeRun(this Action action)
{
    if (PlatformProvider.Current.PropertyChangeNotificationsOnUIThread)
    {
        OnUIThread(action);
    }
    else
    {
        action();
    }
}

```

src/Caliburn.Micro.Platform.Core/RegExHelper.cs
```diff
@@ -40,12 +91,12 @@ public static string GetCaptureGroup(string groupName, string regEx)
         /// <returns>Namespace converted to a regular expression</returns>
         public static string NamespaceToRegEx(string srcNamespace)
         {
-            //Need to escape the "." as it's a special character in regular expression syntax
-            var nsencoded = srcNamespace.Replace(".", @"\.");
+            // Need to escape the "." as it's a special character in regular expression syntax
+            var namespaceEncoded = srcNamespace.Replace(".", @"\.");
```
KasperSK, line 95: If we change from nsencoded I think yo should also change srcNamespace.

KasperSK (changes requested):
Overall it looks good, I have just two minor changes.

src/Caliburn.Micro.Core/DebugLog.cs
```diff
@@ -6,7 +6,7 @@
 namespace Caliburn.Micro
 {
     /// <summary>
-    ///   A simple logger thats logs everything to the debugger.
```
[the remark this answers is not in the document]
K4PS3: should I push another commit with these changes or can you add them from your side?

src/Caliburn.Micro.Platform.Core/RegExHelper.cs
K4PS3: Local variables should be descriptive, and their length does not matter, as they have a narrow scope. The name can also be namespaceEscaped, as escaping characters in text is quite different from encoding them.

K4PS3:
@vb2ae @KasperSK 
I just pushed another commit with these changes:
- Moved 'SafeRun' to 'Execute' Helper class.
- Changed all usage of 'SafeRun' to use the new location of the method.
- Removed the "'s" from the documentation of DebugLog class.
- Changed the local variable name from (namespaceEncoded) to 'encodedNamespace'.

@KasperSK Changing the parameter name from 'srcNamespace' to 'sourceNamespace' will be considered 'Breaking change' and maybe affect someone using it, because it's the public API surface.
Ex. any call with argument name like this will break:
RegExHelper.NamespaceToRegEx(srcNamespace: nsSource + ".")
Are we good to go with this change?

src/Caliburn.Micro.Core/DebugLog.cs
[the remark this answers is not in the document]
K4PS3: @KasperSK Done with the [second commit](https://github.com/Caliburn-Micro/Caliburn.Micro/pull/963#issuecomment-2611565473) in this pull request.

src/Caliburn.Micro.Platform.Core/RegExHelper.cs
K4PS3: @KasperSK can you please take a look at [the new commit notes](https://github.com/Caliburn-Micro/Caliburn.Micro/pull/963#issuecomment-2611565473).

KasperSK:
> @vb2ae @KasperSK I just pushed another commit with these changes:
> 
> * Moved 'SafeRun' to 'Execute' Helper class.
> * Changed all usage of 'SafeRun' to use the new location of the method.
> * Removed the "'s" from the documentation of DebugLog class.
> * Changed the local variable name from (namespaceEncoded) to 'encodedNamespace'.
> 
> @KasperSK Changing the parameter name from 'srcNamespace' to 'sourceNamespace' will be considered 'Breaking change' and maybe affect someone using it, because it's the public API surface. Ex. any call with argument name like this will break: RegExHelper.NamespaceToRegEx(srcNamespace: nsSource + ".") Are we good to go with this change?

Regarding the `SafeRun` method could we give it a more descriptive name? To me safe indicates that nothing can go wrong (Exceptions) so I think a more fitting name would be `DispatchIfRequired`  or something along those lines. 

We can do that name change since the current major version is 4 and we are going to release this with version 5. The reason I brought it up was that it felt mismatched to have one name use shorthand notation and not the other.

vb2ae (approved):
SafeRun makes the code look cleaner.  Documentation is always good to have

KasperSK:
> SafeRun makes the code look cleaner. Documentation is always good to have

Yeah, but I still think we should rename it having a public Method called SafeRun implies that nothing can go wrong and I don't think that is the case here as all it does is dispatch to the UI thread if the platform requires it.

KasperSK (changes requested):

PR #994, opened 2025-07-24
Update WindowManager.cs

vb2ae:
This pull request updates the logic for setting the title of a window in the `CreateWindowAsync` method to handle a specific edge case. The change ensures that windows with a default title of "Window" are treated as if they have no title and are updated accordingly.

### Enhancements to window title handling:
* [`src/Caliburn.Micro.Platform/Platforms/netcore-avalonia/WindowManager.cs`](diffhunk://#diff-cb9fc9540a877be057da1b0e2d493e8c12c6222b80b97ae9e282e3fe4f2f2527L140-R140): Modified the condition in the `CreateWindowAsync` method to check if the window's title is either empty or set to the default value "Window" before applying a binding to the `DisplayName` property. This ensures better handling of default window titles.

### DIFF
--- a/src/Caliburn.Micro.Platform/Platforms/netcore-avalonia/WindowManager.cs
+++ b/src/Caliburn.Micro.Platform/Platforms/netcore-avalonia/WindowManager.cs
@@ -137,11 +137,11 @@ namespace Caliburn.Micro
             var view = EnsureWindow(rootModel, view1);
 
             var haveDisplayName = rootModel as IHaveDisplayName;
-            if (string.IsNullOrEmpty(view.Title) && haveDisplayName != null && !ConventionManager.HasBinding(view, Window.TitleProperty))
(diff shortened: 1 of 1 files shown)


KasperSK:
> Can the pre-release version of Caliburn.Micro.Avalonia be updated on Nuget?

Sure we can do that.

869570967:
Hello, it seems that Caliburn.Micro.Platform.Core has not been updated to version 5.0.231-beta, which has caused issues with updating the Caliburn.Micro.Avalonia library that depends on it. @KasperSK

KasperSK:
Actually Caliburn.Micro.Platform.Core should not be exposed as a package. Let me look into that.

869570967:
Hello, do you have any results yet? I am quite anxious on my side because there are several key bugs in the previous 181-bate version, and I urgently need to update and test. @KasperSK

